PR #2, opened 2017-05-23
Added rules and build scripts

dsgouda:
Replaces https://github.com/Azure/azure-open-api-validator/pull/1


sergey-shandar:
https://gist.github.com/sergey-shandar/f175b4343dfbe923263275d6181f0dfb

dsgouda:
@olydis this is the very first PR I wish to merge.

.travis.yml
```diff
@@ -0,0 +1,9 @@
+language: node
+dist: trusty
+language: node_js
+node_js:
+  - "6.9.5"
+before_script:
+  - npm install
```
vishrutshah: Are you sure you need the `npm install` ? I believe you should not, please double check

vishrutshah (changes requested):

.travis.yml
dsgouda: Referred [this](https://docs.travis-ci.com/user/languages/javascript-with-nodejs/) what is the problem with this?

src/azure-openapi-validator/tests/composite-azure-test.ts
```diff
@@ -0,0 +1,23 @@
...
+import { AutoRestPluginHost } from "../../jsonrpc/plugin-host";
+import { run } from "../../azure-openapi-validator";
+import {
+  assertValidationRuleCount,
+  collectTestMessagesFromValidator
+} from './utilities/tests-helper';
+import { DescriptionMustNotBeNodeName } from '../rules/DescriptionMustNotBeNodeName';
+
+@suite class CompositeAzureTests {
+  @test @timeout(120000) async "description should not be parameter name"() {
```
olydis: note that you can run mocha with a default timeout - consider changing, say, `gulp test` to use a default of 2min instead of restating it in every test :) I'd expect `@timeout` decorators to be the exception

olydis (approved):
nice!

src/azure-openapi-validator/rule.ts
```diff
@@ -0,0 +1,36 @@
...
+interface ValidationMessage {
+  message: string;
+  location: JsonPath;
+}
+
+export interface Rule {
+  readonly id: string; // see Mxxx codes on https://github.com/Azure/azure-rest-api-specs/blob/master/documentation/openapi-authoring-automated-guidelines.md
+  readonly name: string; // see same website as above
+  readonly category: ("RPCViolation" | "OneAPIViolation" | "SDKViolation")[];
+  readonly severity: "error" | "warning";
```
veronicagg: is there a reason for error/warning not to start with upper case?

src/azure-openapi-validator/rule.ts
```diff
@@ -0,0 +1,36 @@
...
+  "composed"
+}
+
+interface ValidationMessage {
+  message: string;
+  location: JsonPath;
+}
+
+export interface Rule {
+  readonly id: string; // see Mxxx codes on https://github.com/Azure/azure-rest-api-specs/blob/master/documentation/openapi-authoring-automated-guidelines.md
```
veronicagg: in comment Mxxx, may need to turn into Rxxx

src/azure-openapi-validator/rules/ControlCharactersAreNotAllowed.ts
```diff
@@ -0,0 +1,32 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+import { MergeStates, OpenApiTypes, rules } from '../rule';
+export const ControlCharactersAreNotAllowed: string = "ControlCharactersAreNotAllowed";
+
+rules.push({
+  id: "SE40",
```
veronicagg: what would happen if we miss one of the properties during the implementation of another rule? are there defaults?

src/azure-openapi-validator/rules/DescriptionMustNotBeNodeName.ts
```diff
@@ -0,0 +1,39 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+import { MergeStates, OpenApiTypes, rules } from '../rule';
+export const DescriptionMustNotBeNodeName: string = "DescriptionMustNotBeNodeName";
+rules.push({
+  id: "M2065",
```
veronicagg: id update?

src/azure-openapi-validator/rule.ts
```diff
@@ -0,0 +1,36 @@
...
+  readonly name: string; // see same website as above
+  readonly category: ("RPCViolation" | "OneAPIViolation" | "SDKViolation")[];
```
veronicagg: do we need category to be an array? what's a case where a rule could fall into 2?

src/azure-openapi-validator/rules/DescriptionMustNotBeNodeName.ts
```diff
@@ -0,0 +1,39 @@
...
+  name: DescriptionMustNotBeNodeName,
+  severity: "error",
+  category: ["RPCViolation"],
+  mergeState: MergeStates.composed,
+  openapiType: OpenApiTypes.arm,
+
+  appliesTo_JsonQuery: "$..*[?(@.description)]",
+  run: function* (doc, node, path) {
+    const msg: string = "Description must not match the name of the node it is supposed to describe";
```
veronicagg: do we need a '.' at the end?

src/azure-openapi-validator/rules/DescriptionMustNotBeNodeName.ts
```diff
@@ -0,0 +1,39 @@
...
+        return;
+      }
+      if (node['name'].toLowerCase() === TrimDescription(node.description)) {
+        yield { message: `${msg} Node name:'${node.name}' Description:'${node.Description}'`, location: path.concat(['description']) };
+      }
+    }
+    else if (nodeName.toLowerCase() === TrimDescription(node.description)) {
+      yield { message: `${msg} Node name:'${nodeName}' Description: '${node.description}'`, location: path.concat(['description']) };
+    } else if (TrimDescription(node.description) === 'description') {
+      yield { message: "Description cannot be named as 'Description'", location: path.concat(['description']) };
```
veronicagg: "." at the end?

src/azure-openapi-validator/tests/composite-azure-test.ts
```diff
@@ -0,0 +1,23 @@
...
+import { AutoRestPluginHost } from "../../jsonrpc/plugin-host";
+import { run } from "../../azure-openapi-validator";
+import {
+  assertValidationRuleCount,
+  collectTestMessagesFromValidator
+} from './utilities/tests-helper';
+import { DescriptionMustNotBeNodeName } from '../rules/DescriptionMustNotBeNodeName';
+
+@suite class CompositeAzureTests {
+  @test async "description should not be parameter name"() {
```
veronicagg: is it parameter or property? I wonder since the file name says DescriptionSameAsPropertyName

src/azure-openapi-validator/tests/resources/DescriptionSameAsPropertyName.json
```diff
@@ -0,0 +1,116 @@
...
+                "parameters": [
+                    {
+                        "name": "resourceGroupName",
+                        "in": "path",
+                        "required": true,
+                        "type": "string",
+                        "description": "The name of the resource group."
+                    },
+                    {
+                        "name": "name",
```
veronicagg: should we name this more explicitly like NodeNameSameAsDescription or similar?

veronicagg (changes requested):
Few comments, rest looks good to me.

.travis.yml
vishrutshah: I think when you say your project if of type `node_js` then travis will understand that you need to do `npm install`. It'll do it on your behalf and you don't need to be explicit about it by `before_script` section. 

As you see that Travis does perform this twice. It's unnecessary. https://travis-ci.org/Azure/azure-openapi-validator/builds/237738002#L139 & https://travis-ci.org/Azure/azure-openapi-validator/builds/237738002#L496

src/azure-openapi-validator/rule.ts
olydis: at least that's the format expected by the AutoRest protocol right [here](https://github.com/Azure/azure-openapi-validator/pull/2/files#diff-4ef9d866ac732d9681e4ce81dab0f934R30), so if spelling is changed here it has to be mapped back when sending the message... potentially unnecessary complexity

src/azure-openapi-validator/rules/ControlCharactersAreNotAllowed.ts
olydis: > what would happen 

it won't compile

> are there defaults?

no, but would that make sense?

src/azure-openapi-validator/rule.ts
dsgouda: What @olydis said

src/azure-openapi-validator/rules/ControlCharactersAreNotAllowed.ts
dsgouda: No defaults won't make sense since we really want authors to definitely provide these details for a validation rule.

src/azure-openapi-validator/rule.ts
dsgouda: Good point, this should be exclusive.

src/azure-openapi-validator/rule.ts
dsgouda: I took a look at https://github.com/Azure/azure-rest-api-specs/blob/9b50c0eed8704ad714a21b5a9cc876b00dbf9d11/documentation/openapi-authoring-automated-guidelines.md
Looks like we are still using ```Mxxx``` while the rules have been updated to ```S/Rxxx```
Where do we add new ids for rules. @kirthik

vishrutshah (approved):

veronicagg (approved):

vishrutshah (changes requested):
Just saw that we are checking in binary file to Git `azure-openapi-validator-0.1.0.tgz` Is it intentional ?

dsgouda:
Nope, fixed.

vishrutshah (approved):

PR #8, opened 2017-06-01
Validating items property must exist for an array type

dsgouda:


src/azure-openapi-linter/rules/ArraySchemaMustHaveItems.ts
```diff
@@ -0,0 +1,31 @@
...
+ *--------------------------------------------------------------------------------------------*/
+import { MergeStates, OpenApiTypes, rules } from '../rule';
+export const ArraySchemaMustHaveItems: string = "ArraySchemaMustHaveItems";
+
+const jp = require('jsonpath');
+
+rules.push({
+  id: "S2009",
+  name: ArraySchemaMustHaveItems,
+  severity: "warning",
```
veronicagg: is this the violation that would result in AutoRest producing a code generation error?

src/azure-openapi-linter/rules/ArraySchemaMustHaveItems.ts
```diff
@@ -0,0 +1,31 @@
...
+  category: "SDKViolation",
+  mergeState: MergeStates.individual,
+  openapiType: OpenApiTypes.arm,
+  appliesTo_JsonQuery: "$.definitions.*.properties.*.type",
+  run: function* (doc, node, path) {
+    const msg: string = "Please provide an items property for array type: ";
```
veronicagg: should 'items' be in quotes?

src/azure-openapi-linter/tests/resources/ArraySchemaWithoutItems.json
```diff
@@ -0,0 +1,77 @@
...
+            "description": "test api version"
+        }
+    },
+    "definitions": {
+        "Models": {
+            "description": "the models to return",
+            "properties": {
+                "mymodels": {
+                    "type": "array",
+                    "items": {
```
veronicagg: what's the violation of the rule for this model? isn't "items" defined for the array type?

src/azure-openapi-linter/rules/ArraySchemaMustHaveItems.ts
```diff
@@ -0,0 +1,31 @@
...
+    if (node === 'array') {
+      // climb up the json path to get hold of the parent element
+      // check if the parent element has an items node defined
+      const propObject = jp.query(doc, path.splice(0, path.length - 2).join('.'));
```
veronicagg: why path.length-2? it'd be good to get a comment before this line with an example of what we're looking to accomplish, so we can interpret the code more easily.

src/azure-openapi-linter/rules/ArraySchemaMustHaveItems.ts
dsgouda: That's right.

src/azure-openapi-linter/tests/resources/ArraySchemaWithoutItems.json
dsgouda: Good catch, almost missed that, fixed the logic and the test case.

src/azure-openapi-linter/rules/ArraySchemaMustHaveItems.ts
veronicagg: If so, should it be severity: error?

src/azure-openapi-linter/rules/ArraySchemaMustHaveItems.ts
```diff
@@ -12,19 +12,14 @@ rules.push({
   name: ArraySchemaMustHaveItems,
   severity: "warning",
   category: "SDKViolation",
-  mergeState: MergeStates.individual,
+  mergeState: MergeStates.composed,
```
veronicagg: why should this one be composed instead of individual? just trying to understand the difference in this case.

src/azure-openapi-linter/rules/ArraySchemaMustHaveItems.ts
```diff
@@ -12,19 +12,14 @@ rules.push({
...
   openapiType: OpenApiTypes.arm,
-  appliesTo_JsonQuery: "$.definitions.*.properties.*.type",
+  appliesTo_JsonQuery: "$.definitions.*.properties[?(@.type)]",
```
veronicagg: could you provide more info on how we're building these query expressions?

src/azure-openapi-linter/tests/resources/ArraySchemaWithoutItems.json
```diff
@@ -57,10 +57,7 @@
             "description": "the models to return",
             "properties": {
                 "mymodels": {
-                    "type": "array",
-                    "items": {
-                        "$ref": "#/definitions/Model"
-                    }
+                    "type": "array"
```
veronicagg: if this was incorrect, how was the test passing? why were we not catching the fact that the spec was not reporting a violation?

dsgouda:
@veronicagg  I feel it should be error since AutoRest doesn't exactly handle the resultant autogen error gracefully enough, I can update this if you agree

src/azure-openapi-linter/rules/ArraySchemaMustHaveItems.ts
dsgouda: The info is provided in readme docs, a simple tutorial is available [here](https://www.npmjs.com/package/jsonpath)

src/azure-openapi-linter/rules/ArraySchemaMustHaveItems.ts
dsgouda: Either should do really since this works irrespective of whether the spec is in merged state or not.

src/azure-openapi-linter/tests/resources/ArraySchemaWithoutItems.json
dsgouda: Fixed

src/azure-openapi-linter/rules/ArraySchemaMustHaveItems.ts
```diff
@@ -0,0 +1,24 @@
...
+  name: ArraySchemaMustHaveItems,
+  severity: "error",
+  category: "SDKViolation",
+  mergeState: MergeStates.composed,
+  openapiType: OpenApiTypes.arm,
+  appliesTo_JsonQuery: "$.definitions.*.properties[?(@.type==='array')]",
+  run: function* (doc, node, path) {
+    const msg: string = "Please provide an 'items' property for array type: ";
+    if (!node.hasOwnProperty('items')) {
+      yield { message: `${msg} '${path[path.length - 2]}'`, location: path.slice(0, path.length - 2) };
```
veronicagg: is path-length - 2 giving back the name of the model? just double checking if that's the return or if it's walks up one more level.

dsgouda: It gives the name of the model, one level up would be the `properties` node.

PR #13, opened 2017-06-08
Renaming package to azure-openapi-validator

dsgouda:
Per suggestions by @olydis, renaming the repo and published package to `@microsoft.azure/azure-openapi-validator`


package.json
```diff
@@ -1,14 +1,14 @@
 {
-  "name": "azure-openapi-linter",
-  "version": "0.1.4",
-  "description": "Azure OpenAPI Linter",
+  "name": "@microsoft.azure/azure-openapi-validator",
+  "version": "0.1.5",
```
vishrutshah: Why do you want to start a brand new package from 0.1.5. I'd go for 0.1.0 again

package.json
```diff
@@ -1,14 +1,14 @@
 {
-  "name": "azure-openapi-linter",
-  "version": "0.1.4",
-  "description": "Azure OpenAPI Linter",
+  "name": "@microsoft.azure/azure-openapi-validator",
```
vishrutshah: I am not sure how this works but is this intentional change?

dsgouda: Yes, quite intentional, @fearthecowboy renamed all azure-related npm packages to read as ```@microsoft.azure/*``` for consistency. Perhaps we could rename some of our other npm packages accordingly.

vishrutshah: As per our offline conversation, this is not a stand alone package and AutoRest would depend on this package instead of other way around. So we agreed to name it as `@microsoft.azure/` and make it not discoverable independently.

Issue #36, opened 2017-07-12
[BUG] Top level properties should not be repeated inside properties bag


veronicagg:
Please refer to details at https://github.com/Azure/autorest/issues/2304

sarangan12:
Task completed

PR #79, opened 2017-07-28
Code changes to fix bug in arm resources properties bag rule

sarangan12:
Related to https://github.com/Azure/azure-openapi-validator/issues/36 

This rule has the following issues:
1. The rule ignores the reference values
2. The 'tags' property is misspelled as 'tag'
3. The rule ignores the allOf values

I have covered the missing cases and added possible examples. Please review and approve

### DIFF
--- a/src/dotnet/OpenAPI.Validator.Tests/OpenAPIModelerValidationTests.cs
+++ b/src/dotnet/OpenAPI.Validator.Tests/OpenAPIModelerValidationTests.cs
@@ -200,6 +200,37 @@ namespace OpenAPI.Validator.Tests
         }
 
         [Fact]
+        public void ArmResourcePropertiesBagMultipleViolationsValidation()
+        {
+            var messages = GetValidationMessagesForRule<ArmResourcePropertiesBag>("arm-resource-properties-bag-multiple-violations.json");
+            Assert.Equal(messages.Count(), 1);
+            Assert.Equal(messages.First().Message.Contains("[name, type]"), true);
+        }
+
+        [Fact]
+        public void ArmResourcePropertiesBagMultipleLevelViolationsValidation()
+        {
+            var messages = GetValidationMessagesForRule<ArmResourcePropertiesBag>("arm-resource-properties-bag-multiple-level-violations.json");
+            Assert.Equal(messages.Count(), 2);
+            Assert.Equal(messages.First().Message.Contains("[name, type]"), true);
+            Assert.Equal(messages.Last().Message.Contains("[location, id]"), true);
+        }
+
+        [Fact]
+        public void ArmResourcePropertiesBagWithReferenceValidation()
+        {
+            var messages = GetValidationMessagesForRule<ArmResourcePropertiesBag>("arm-resource-properties-bag-with-reference.json");
+            Assert.Equal(messages.Count(), 1);
+        }
+
+        [Fact]
+        public void ArmResourcePropertiesBagWithMultipleLevelReferenceValidation()
+        {
+            var messages = GetValidationMessagesForRule<ArmResourcePropertiesBag>("arm-resource-properties-bag-with-multiple-level-reference.json");
+            Assert.Equal(messages.Count(), 1);
+        }
+
+        [Fact]
         public void CollectionObjectsPropertiesNamingValidation()
         {
             var messages = GetValidationMessagesForRule<CollectionObjectPropertiesNaming>("collection-objects-naming.json");
--- a/src/dotnet/OpenAPI.Validator/Validation/ArmResourcePropertiesBag.cs
+++ b/src/dotnet/OpenAPI.Validator/Validation/ArmResourcePropertiesBag.cs
@@ -7,6 +7,7 @@ using OpenAPI.Validator.Properties;
 using System.Collections.Generic;
 using OpenAPI.Validator.Model;
 using System.Linq;
+using OpenAPI.Validator;
 
 namespace OpenAPI.Validator.Validation
 {
@@ -67,25 +68,35 @@ namespace OpenAPI.Validator.Validation
             }
         }
 
-        private string GetName(string reference)
-        {
-            return reference.Substring(reference.LastIndexOf("/") + 1);
-        }
-
+        /// <summary>
+        /// Method to check the model for violation of this rule.
+        /// This method checks if the model has propeties bag. If so and it
+        /// has reference then the reference is processed. Else, the individual
+        /// properties are checked for violation.
+        ///
+        /// Then, if the model has allOf's, they need to be processed individually.
+        /// </summary>
+        /// <param name="resourceModel">schema of the model</param>
+        /// <param name="resourceModelName">name of the model</param>
+        /// <param name="definitions">definitions</param>
+        /// <param name="context">context</param>
+        /// <param name="violations">violations table</param>
         private void CheckModelForViolation(Schema resourceModel, string resourceModelName, Dictionary<string, Schema> definitions, RuleContext context, Dictionary<string, IEnumerable<string>> violations)
         {
             if (resourceModel.Properties?.ContainsKey("properties") == true)
             {
                 string referenceName = resourceModel.Properties["properties"].Reference;
-                if (referenceName != null)
+                if (!string.IsNullOrWhiteSpace(referenceName))
                 {
-                    CheckModelForViolation(Schema.FindReferencedSchema(referenceName, definitions), GetName(referenceName), definitions, context, violations);
+                    CheckModelForViolation(Schema.FindReferencedSchema(referenceName, definitions), Validator.Extensions.StripDefinitionPath(referenceName), definitions, context, violations);
                 }
-
-                IEnumerable<string> violatingProperties = resourceModel.Properties["properties"].Properties?.Keys?.Intersect(ArmPropertiesBag);
-                if (violatingProperties?.Any() == true)
+                else
                 {
-                    violations[resourceModelName] = violatingProperties;
+                    IEnumerable<string> violatingProperties = resourceModel.Properties["properties"].Properties?.Keys?.Intersect(ArmPropertiesBag);
+                    if (violatingProperties?.Any() == true)
+                    {
+                        violations[resourceModelName] = violatingProperties;
+                    }
                 }
             }
 
@@ -93,7 +104,7 @@ namespace OpenAPI.Validator.Validation
             {
                 foreach (Schema schema in resourceModel.AllOf)
                 {
-                    CheckModelForViolation(Schema.FindReferencedSchema(schema.Reference, definitions), GetName(schema.Reference), definitions, context, violations);
+                    CheckModelForViolation(Schema.FindReferencedSchema(schema.Reference, definitions), Validator.Extensions.StripDefinitionPath(schema.Reference), definitions, context, violations);
                 }
             }
         }
--- a/src/dotnet/OpenAPI.Validator/Validation/ArmResourcePropertiesBag.cs
+++ b/src/dotnet/OpenAPI.Validator/Validation/ArmResourcePropertiesBag.cs
@@ -95,7 +95,7 @@ namespace OpenAPI.Validator.Validation
                     else
                     {
                         IEnumerable<string> violatingProperties = resourceModel.Properties["properties"].Properties?.Keys?.Intersect(ArmPropertiesBag);
-                        if (violatingProperties.Any() == true)
+                        if (violatingProperties?.Any() == true)
                         {
                             violations[resourceModelName] = violatingProperties;
                         }
--- a/src/dotnet/OpenAPI.Validator/Validation/ArmResourcePropertiesBag.cs
+++ b/src/dotnet/OpenAPI.Validator/Validation/ArmResourcePropertiesBag.cs
@@ -83,28 +83,31 @@ namespace OpenAPI.Validator.Validation
         /// <param name="violations">violations table</param>
         private void CheckModelForViolation(Schema resourceModel, string resourceModelName, Dictionary<string, Schema> definitions, RuleContext context, Dictionary<string, IEnumerable<string>> violations)
         {
-            if (resourceModel.Properties?.ContainsKey("properties") == true)
+            if(resourceModel != null)
             {
-                string referenceName = resourceModel.Properties["properties"].Reference;
-                if (!string.IsNullOrWhiteSpace(referenceName))
+                if (resourceModel.Properties?.ContainsKey("properties") == true)
                 {
-                    CheckModelForViolation(Schema.FindReferencedSchema(referenceName, definitions), Validator.Extensions.StripDefinitionPath(referenceName), definitions, context, violations);
-                }
-                else
-                {
-                    IEnumerable<string> violatingProperties = resourceModel.Properties["properties"].Properties?.Keys?.Intersect(ArmPropertiesBag);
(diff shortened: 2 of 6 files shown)


src/dotnet/OpenAPI.Validator/Validation/ArmResourcePropertiesBag.cs
```diff
@@ -47,19 +47,54 @@ public class ArmResourcePropertiesBag : TypedRule<Dictionary<string, Schema>>
...
+        }
+
+        private void CheckModelForViolation(Schema resourceModel, string resourceModelName, Dictionary<string, Schema> definitions, RuleContext context, Dictionary<string, IEnumerable<string>> violations)
+        {
+            if (resourceModel.Properties?.ContainsKey("properties") == true)
+            {
+                string referenceName = resourceModel.Properties["properties"].Reference;
+                if (referenceName != null)
+                {
+                    CheckModelForViolation(Schema.FindReferencedSchema(referenceName, definitions), GetName(referenceName), definitions, context, violations);
```
vishrutshah: This is a recursive method. Does is work when we have recursive models? Did you get chance to verify that?

src/dotnet/OpenAPI.Validator/Validation/ArmResourcePropertiesBag.cs
```diff
@@ -47,19 +47,54 @@ public class ArmResourcePropertiesBag : TypedRule<Dictionary<string, Schema>>
...
+        {
+            return reference.Substring(reference.LastIndexOf("/") + 1);
+        }
+
+        private void CheckModelForViolation(Schema resourceModel, string resourceModelName, Dictionary<string, Schema> definitions, RuleContext context, Dictionary<string, IEnumerable<string>> violations)
+        {
+            if (resourceModel.Properties?.ContainsKey("properties") == true)
+            {
+                string referenceName = resourceModel.Properties["properties"].Reference;
+                if (referenceName != null)
```
vishrutshah: Would this be better check - `IsNullOrWhiteSpace` ?

src/dotnet/OpenAPI.Validator/Validation/ArmResourcePropertiesBag.cs
```diff
@@ -47,19 +47,54 @@ public class ArmResourcePropertiesBag : TypedRule<Dictionary<string, Schema>>
...
+                }
+            }
+        }
+
+        private string GetName(string reference)
+        {
+            return reference.Substring(reference.LastIndexOf("/") + 1);
+        }
+
+        private void CheckModelForViolation(Schema resourceModel, string resourceModelName, Dictionary<string, Schema> definitions, RuleContext context, Dictionary<string, IEnumerable<string>> violations)
```
vishrutshah: Should we check for `throw new ArgumentNullException(nameof(resourceModel))` ?

src/dotnet/OpenAPI.Validator/Validation/ArmResourcePropertiesBag.cs
```diff
@@ -47,19 +47,54 @@ public class ArmResourcePropertiesBag : TypedRule<Dictionary<string, Schema>>
...
+        {
```
vishrutshah: Should we check for `resourceModelName`'s null or empty string?

src/dotnet/OpenAPI.Validator/Validation/ArmResourcePropertiesBag.cs
```diff
@@ -47,19 +47,54 @@ public class ArmResourcePropertiesBag : TypedRule<Dictionary<string, Schema>>
...
+                string referenceName = resourceModel.Properties["properties"].Reference;
+                if (referenceName != null)
+                {
+                    CheckModelForViolation(Schema.FindReferencedSchema(referenceName, definitions), GetName(referenceName), definitions, context, violations);
+                }
+
+                IEnumerable<string> violatingProperties = resourceModel.Properties["properties"].Properties?.Keys?.Intersect(ArmPropertiesBag);
+                if (violatingProperties?.Any() == true)
+                {
+                    violations[resourceModelName] = violatingProperties;
```
vishrutshah: Is it acceptable to overwrite the violations in case `resourceModelName` already existed in the `violations` dictionary?

src/dotnet/OpenAPI.Validator.Tests/OpenAPIModelerValidationTests.cs
```diff
@@ -200,6 +200,37 @@ public void ArmResourcePropertiesBagValidation()
...
+            Assert.Equal(messages.Count(), 2);
+            Assert.Equal(messages.First().Message.Contains("[name, type]"), true);
+            Assert.Equal(messages.Last().Message.Contains("[location, id]"), true);
+        }
+
+        [Fact]
+        public void ArmResourcePropertiesBagWithReferenceValidation()
+        {
+            var messages = GetValidationMessagesForRule<ArmResourcePropertiesBag>("arm-resource-properties-bag-with-reference.json");
+            Assert.Equal(messages.Count(), 1);
```
vishrutshah: Just curious: Any reason for not validating the Message.Contains like above test cases?

src/dotnet/OpenAPI.Validator.Tests/OpenAPIModelerValidationTests.cs
```diff
@@ -200,6 +200,37 @@ public void ArmResourcePropertiesBagValidation()
...
+        }
+
+        [Fact]
+        public void ArmResourcePropertiesBagWithMultipleLevelReferenceValidation()
+        {
+            var messages = GetValidationMessagesForRule<ArmResourcePropertiesBag>("arm-resource-properties-bag-with-multiple-level-reference.json");
+            Assert.Equal(messages.Count(), 1);
```
vishrutshah: Just curious: Any reason for not validating the Message.Contains like above test cases?

src/dotnet/OpenAPI.Validator/Validation/ArmResourcePropertiesBag.cs
```diff
@@ -47,19 +47,54 @@ public class ArmResourcePropertiesBag : TypedRule<Dictionary<string, Schema>>
...
+                foreach (string modelName in violations.Keys)
+                {
+                    yield return new ValidationMessage(new FileObjectPath(context.File,
+                    context.Path.AppendProperty(modelName).AppendProperty("properties")), this, modelName,
+                                                   string.Join(", ", violations[modelName]));
+                }
+            }
+        }
+
+        private string GetName(string reference)
```
veronicagg: should we be using StripDefinitionPath method?: https://github.com/Azure/azure-openapi-validator/blob/42ae8eedab0a3894443966e4513fd28ae5aa77d8/src/dotnet/OpenAPI.Validator/Extensions.cs#L96

src/dotnet/OpenAPI.Validator/Validation/ArmResourcePropertiesBag.cs
```diff
@@ -47,19 +47,54 @@ public class ArmResourcePropertiesBag : TypedRule<Dictionary<string, Schema>>
...
-            foreach (var violatingModel in violatingModels)
+            IEnumerable<string> resourceModelsWithPropertiesBag = context.ResourceModels.Where(resourceModel => definitions[resourceModel].Properties?.ContainsKey("properties") == true);
+            foreach(string resourceModelName in resourceModelsWithPropertiesBag)
             {
-                var violatingProperties = definitions[violatingModel].Properties.Keys.Union(definitions[violatingModel].Properties["properties"].Properties.Keys).Intersect(ArmPropertiesBag);
-                yield return new ValidationMessage(new FileObjectPath(context.File, context.Path.AppendProperty(violatingModel).AppendProperty("properties")), this, violatingModel,
-                                                   string.Join(", ", violatingProperties));
+                Dictionary<string, IEnumerable<string>> violations = new Dictionary<string, IEnumerable<string>>();             
+                CheckModelForViolation(definitions[resourceModelName], resourceModelName, definitions, context, violations);
```
veronicagg: would this method "EnumerateProperties" https://github.com/Azure/azure-openapi-validator/blob/42ae8eedab0a3894443966e4513fd28ae5aa77d8/src/dotnet/OpenAPI.Validator/Model/Utilities/ValidationUtilities.cs#L109 return all the properties we'd need to compare against?
There's also this one https://github.com/Azure/azure-openapi-validator/blob/42ae8eedab0a3894443966e4513fd28ae5aa77d8/src/dotnet/OpenAPI.Validator/Model/Utilities/ValidationUtilities.cs#L88 for allOfs, though you may not need it. 
Let's just make sure those methods would return top level properties which are the ones needed here.

src/dotnet/OpenAPI.Validator.Tests/OpenAPIModelerValidationTests.cs
sarangan12: Same as above

src/dotnet/OpenAPI.Validator/Validation/ArmResourcePropertiesBag.cs
sarangan12: The logic is that for each model there will be one entry in the violations table.

src/dotnet/OpenAPI.Validator/Validation/ArmResourcePropertiesBag.cs
sarangan12: Yes. The test case covers those scenarios also.

src/dotnet/OpenAPI.Validator/Validation/ArmResourcePropertiesBag.cs
sarangan12: Done

src/dotnet/OpenAPI.Validator/Validation/ArmResourcePropertiesBag.cs
sarangan12: The method has been removed

sarangan12: I was not aware of this method. Used this with the latest code. Thanks

src/dotnet/OpenAPI.Validator/Validation/ArmResourcePropertiesBag.cs
```diff
@@ -47,19 +47,54 @@ public class ArmResourcePropertiesBag : TypedRule<Dictionary<string, Schema>>
...
+                {
+                    yield return new ValidationMessage(new FileObjectPath(context.File,
+                    context.Path.AppendProperty(modelName).AppendProperty("properties")), this, modelName,
+                                                   string.Join(", ", violations[modelName]));
+                }
+            }
+        }
+
+        private string GetName(string reference)
+        {
```
sarangan12: The method has been removed

src/dotnet/OpenAPI.Validator/Validation/ArmResourcePropertiesBag.cs
sarangan12: I dont think so

src/dotnet/OpenAPI.Validator/Validation/ArmResourcePropertiesBag.cs
sarangan12: I do not want to enumerate the properties. Instead, I want to point to the exact location of the violation. For example, if M1 is the model that I am analyzing which does an allOf on M2 (Which actually has a violation) then the message should point to M2 and not M1. So, I am not using the enumerate properties

sarangan12:
@vishrutshah @veronicagg Responded to your comments. Please review

src/dotnet/OpenAPI.Validator/Validation/ArmResourcePropertiesBag.cs
```diff
@@ -47,19 +48,64 @@ public class ArmResourcePropertiesBag : TypedRule<Dictionary<string, Schema>>
...
+        /// Then, if the model has allOf's, they need to be processed individually.
+        /// </summary>
+        /// <param name="resourceModel">schema of the model</param>
+        /// <param name="resourceModelName">name of the model</param>
+        /// <param name="definitions">definitions</param>
+        /// <param name="context">context</param>
+        /// <param name="violations">violations table</param>
+        private void CheckModelForViolation(Schema resourceModel, string resourceModelName, Dictionary<string, Schema> definitions, RuleContext context, Dictionary<string, IEnumerable<string>> violations)
+        {
+            if (resourceModel.Properties?.ContainsKey("properties") == true)
```
vishrutshah: If i am not mistaken then method `Schema.FindReferencedSchema` returns `null` [here](https://github.com/Azure/azure-openapi-validator/blob/42ae8eedab0a3894443966e4513fd28ae5aa77d8/src/dotnet/OpenAPI.Validator/Model/Schema.cs#L87). We are calling that method at line `108`'s first parameter without checking it for `null`. In that case we need to do a null check here.

In general, I'd always prefer to do a null check for all the arguments of any input function as we don't know how someone is going to call it. To be on the safe side.

src/dotnet/OpenAPI.Validator/Validation/ArmResourcePropertiesBag.cs
```diff
@@ -47,19 +48,64 @@ public class ArmResourcePropertiesBag : TypedRule<Dictionary<string, Schema>>
...
+            {
+                string referenceName = resourceModel.Properties["properties"].Reference;
+                if (!string.IsNullOrWhiteSpace(referenceName))
+                {
+                    CheckModelForViolation(Schema.FindReferencedSchema(referenceName, definitions), Validator.Extensions.StripDefinitionPath(referenceName), definitions, context, violations);
+                }
+                else
+                {
+                    IEnumerable<string> violatingProperties = resourceModel.Properties["properties"].Properties?.Keys?.Intersect(ArmPropertiesBag);
+                    if (violatingProperties?.Any() == true)
```
vishrutshah: Curious: Does `Intersect` ever return `null` ? I thought it doesn't. If it doesn't we can just write 

`if (violatingProperties.Any())`

src/dotnet/OpenAPI.Validator/Validation/ArmResourcePropertiesBag.cs
vishrutshah: Cool. Can you point me to the model that is `circular recursive` as it's had to find in json :)

src/dotnet/OpenAPI.Validator.Tests/OpenAPIModelerValidationTests.cs
vishrutshah: That's good that your test cases are more accurate :) That's why i was wondering that why not follow that here as well :)

src/dotnet/OpenAPI.Validator/Validation/ArmResourcePropertiesBag.cs
sarangan12: Removed it

src/dotnet/OpenAPI.Validator/Validation/ArmResourcePropertiesBag.cs
sarangan12: Added the null check. But it is not required. Because, the null will be returned if the schema is referenced and not available. In that case code model will not be generated at all (errored out) and will not come to this rule. To be on the safe side, added the null check

Issue #44, opened 2017-07-12
ListByResourceGroup and ListBySubcription dont apply to tenant level resources 


veronicagg:
See details as https://github.com/Azure/autorest/issues/2170

sarangan12:
Task completed

PR #80, opened 2017-08-02
Code changes to remove tenant level resources from ListByRG and ListBySubscriptions Rules

sarangan12:
Related to https://github.com/Azure/azure-openapi-validator/issues/44

Please review and approve

### DIFF
--- a/src/dotnet/OpenAPI.Validator/Model/Utilities/ValidationUtilities.cs
+++ b/src/dotnet/OpenAPI.Validator/Model/Utilities/ValidationUtilities.cs
@@ -23,6 +23,36 @@ namespace OpenAPI.Validator.Model.Utilities
         public static readonly Regex ListBySidRegEx = new Regex(@".+_(List|ListBySubscriptionId|ListBySubscription|ListBySubscriptions)$", RegexOptions.IgnoreCase);
         public static readonly Regex ListByRgRegEx = new Regex(@".+_ListByResourceGroup$", RegexOptions.IgnoreCase);
 
+        public static IEnumerable<string> GetTenantResourceModels(IEnumerable<string> resourceModels, ServiceDefinition serviceDefinition)
+        {
+            List<string> tenantResourceModels = new List<string>();
+            if(serviceDefinition.Paths.Any())
+            {
+                foreach (KeyValuePair<string, Dictionary<string, Operation>> path in serviceDefinition.Paths)
+                {
+                    if (!path.Key.Contains("/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/"))
+                    {
+                        // This path does not have the subscriptions and resourcegroups. So, any resource model that is returned
+                        // by a 200 operation could be tagged as a tenant level resource. 
+
+                        string pathValue = path.Key;
+                        Operation getOperation = path.Value.GetValueOrNull("get");
+                        if (getOperation != null && getOperation.Responses?.GetValueOrNull("200") != null)
+                        {
+                            OperationResponse response = getOperation.Responses.GetValueOrNull("200");
+                            string nameToCompare = response.Schema?.Reference?.Substring(response.Schema.Reference.LastIndexOf("/") + 1);
+                            if (!string.IsNullOrEmpty(nameToCompare) && resourceModels.Contains(nameToCompare))
+                            {
+                                tenantResourceModels.Add(nameToCompare);
+                            }
+                        }
+                    }
+                }
+            }
+
+            return tenantResourceModels;
+        }
+
         /// <summary>
         /// Populates a list of 'Resource' models found in the service definition
         /// </summary>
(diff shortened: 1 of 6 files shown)


src/dotnet/OpenAPI.Validator/Model/Utilities/ValidationUtilities.cs
```diff
@@ -23,6 +23,36 @@ public static class ValidationUtilities
...
+                    {
+                        // This path does not have the subscriptions and resourcegroups. So, any resource model that is returned
+                        // by a 200 operation could be tagged as a tenant level resource. 
+
+                        string pathValue = path.Key;
+                        Operation getOperation = path.Value.GetValueOrNull("get");
+                        if (getOperation != null && getOperation.Responses?.GetValueOrNull("200") != null)
+                        {
+                            OperationResponse response = getOperation.Responses.GetValueOrNull("200");
+                            string nameToCompare = response.Schema?.Reference?.Substring(response.Schema.Reference.LastIndexOf("/") + 1);
```
veronicagg: should we be using StripDefinitionPath?

src/dotnet/OpenAPI.Validator/Model/Utilities/ValidationUtilities.cs
```diff
@@ -23,6 +23,36 @@ public static class ValidationUtilities
...
         public static readonly Regex ListByRgRegEx = new Regex(@".+_ListByResourceGroup$", RegexOptions.IgnoreCase);
 
+        public static IEnumerable<string> GetTenantResourceModels(IEnumerable<string> resourceModels, ServiceDefinition serviceDefinition)
+        {
+            List<string> tenantResourceModels = new List<string>();
+            if(serviceDefinition.Paths.Any())
+            {
+                foreach (KeyValuePair<string, Dictionary<string, Operation>> path in serviceDefinition.Paths)
+                {
+                    if (!path.Key.Contains("/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/"))
```
veronicagg: do we need a regex for this? The names in between {} could vary based on what the author of the spec decides to call them as they are parameters, and casing may be different too. 
I believe there are cases in which one or the other may no be there too, should those also be excluded, the description from Gaurav in the issue says "will not have the /resourceGroups or /subscriptions section in the URL", so the "or" makes me wonder.

src/dotnet/OpenAPI.Validator/Model/Utilities/ValidationUtilities.cs
```diff
@@ -23,6 +23,36 @@ public static class ValidationUtilities
         public static readonly Regex ListBySidRegEx = new Regex(@".+_(List|ListBySubscriptionId|ListBySubscription|ListBySubscriptions)$", RegexOptions.IgnoreCase);
         public static readonly Regex ListByRgRegEx = new Regex(@".+_ListByResourceGroup$", RegexOptions.IgnoreCase);
 
+        public static IEnumerable<string> GetTenantResourceModels(IEnumerable<string> resourceModels, ServiceDefinition serviceDefinition)
+        {
+            List<string> tenantResourceModels = new List<string>();
+            if(serviceDefinition.Paths.Any())
+            {
+                foreach (KeyValuePair<string, Dictionary<string, Operation>> path in serviceDefinition.Paths)
```
veronicagg: @ravbhatnagar could you check on this logic, to make sure the lookup of tenant level resource is appropriate?

src/dotnet/OpenAPI.Validator/Model/Utilities/ValidationUtilities.cs
```diff
@@ -23,6 +23,36 @@ public static class ValidationUtilities
...
+                {
+                    if (!path.Key.Contains("/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/"))
+                    {
+                        // This path does not have the subscriptions and resourcegroups. So, any resource model that is returned
+                        // by a 200 operation could be tagged as a tenant level resource. 
+
+                        string pathValue = path.Key;
+                        Operation getOperation = path.Value.GetValueOrNull("get");
```
veronicagg: should we only be considering "get"s in for looking up these resources?

src/dotnet/OpenAPI.Validator/Model/Utilities/ValidationUtilities.cs
sarangan12: Awwwwww... missed it for the second time.. Fixed it. Thanks

src/dotnet/OpenAPI.Validator/Model/Utilities/ValidationUtilities.cs
sarangan12: "get" should be enough

src/dotnet/OpenAPI.Validator/Model/Utilities/ValidationUtilities.cs
sarangan12: regEx would be better. Changed it

ravbhatnagar:
@sarangan12 - Just wanted to make sure if this will be flagged - /subscriptions/{subscriptionId}/providers/Microsoft.Commerce/RateCard ? See the spec - https://github.com/Azure/azure-rest-api-specs/blob/current/specification/commerce/resource-manager/Microsoft.Commerce/2015-06-01-preview/commerce.json
This is not tenant level but still does not have a resource group segment which is not required for this API. This is a subscription wide resource in that sense. So ListBy Resource group violation does not apply.

PR #85, opened 2017-08-08
Case insensitive check for res toplevel props

dsgouda:
From https://github.com/Azure/autorest/pull/2152

### DIFF
--- a/src/dotnet/OpenAPI.Validator.Tests/Resource/OpenAPI/Validation/body-top-level-properties.json
+++ b/src/dotnet/OpenAPI.Validator.Tests/Resource/OpenAPI/Validation/body-top-level-properties.json
@@ -79,7 +79,7 @@
             "modelAsString": false
           }
         },
-        "Location": {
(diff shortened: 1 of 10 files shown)


veronicagg:
@dsgouda I agree with you, having capital letters at the start of the property name is not accepted by other rules, how about "etag" versus "eTag"? maybe we can use something like that in the tests,? so it's not violating the other rule and still helps with coverage?

src/dotnet/OpenAPI.Validator.Tests/Resource/OpenAPI/Validation/body-top-level-properties.json
```diff
@@ -79,10 +79,14 @@
             "modelAsString": false
           }
         },
-        "location": {
+        "Location": {
```
veronicagg: do we still need to make this one and "Type" non compliant with the other rule? could we just leave "eTag" in the example for testing?

PR #87, opened 2017-08-22
Improve the result of GetCamelCasedSuggestion

Porges:
(Moved from https://github.com/Azure/autorest/pull/2548)

When run on our service's definition, `GetCamelCasedSuggestion` returned:

* `nodeAgentSkUid` for `nodeAgentSKUId`
* `inboundNaTpools` for `inboundNATPools`

These are patently silly. This change improves them to:

* `nodeAgentSkuId` for `nodeAgentSKUId`
* `inboundNatPools` for `inboundNATPools`

### DIFF
--- a/src/dotnet/OpenAPI.Validator/Model/Utilities/ValidationUtilities.cs
+++ b/src/dotnet/OpenAPI.Validator/Model/Utilities/ValidationUtilities.cs
@@ -34,7 +34,7 @@ namespace OpenAPI.Validator.Model.Utilities
                     if (!TenantResourceRegEx.IsMatch(path.Key))
                     {
                         // This path does not have the subscriptions and resourcegroups. So, any resource model that is returned
-                        // by a 200 operation could be tagged as a tenant level resource. 
+                        // by a 200 operation could be tagged as a tenant level resource.
 
                         string pathValue = path.Key;
                         Operation getOperation = path.Value.GetValueOrNull("get");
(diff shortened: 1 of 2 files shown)


olydis:
@dsgouda PTAL

src/dotnet/OpenAPI.Validator/Model/Utilities/ValidationUtilities.cs
```diff
@@ -365,33 +365,14 @@ private static IEnumerable<Operation> SelectOperationsFromPaths(string id, Dicti
...
-                    firstUpper = true;
-                    if (char.IsUpper(sb[i]))
-                    {
-                        sb[i] = sb[i].ToString().ToLower()[0];
-                    }
-                }
-            }
-            return sb.ToString();
-        }
+            => Regex.Replace(name, @"(?<=\p{Lu})\p{Lu}+(?=\p{Lu}|\z)", s => s.Value.ToLowerInvariant());
```
dsgouda: I'm all for regexes but even I can't comprehend this on the first go. I'd keep the dirty logic as is since this is bound to evolve further. @veronicagg thoughts?

olydis: note that the logic is *changed* to actually alter the rule, so there's not much to "keep". But I agree, there may be value in rewriting this Regex as code (preferably not dirty, but readable 😄 ) for future maintainability

dsgouda: Ahh, did not notice that, but yes, still see value in turning the regex into code.

veronicagg: Yeah, the reason I haven't reviewed this PR is because I had to set time aside to study the regex :) if we can put it in simpler terms it'd be great, if not, may be an explanation of the regex in comments would help. The current logic suggesting a string is not ideal, but that's the reason we kept it as a suggestion, it was a way to "show" the issue and a compliant option more than having people take the suggestion.

Porges: It's _quite_ clunky if converted to code but I'll comment the regex to explain what it's doing

Porges: ... done that, and added a test for expected behaviour.

Porges:
Thanks!

PR #90, opened 2017-08-26
2.0 ready

olydis:
Makes the extension
- use dotnet 2.0
- compatible with AutoRest 2.0

(I republished the classic validator tagged as 'preview' in order to test, nothing that's out there is affected)


gulpfile.js
```diff
@@ -3,12 +3,26 @@
...
 var gulp = require('gulp');
 var clean = require('gulp-clean');
 var run = require('gulp-run')
 var mocha = require('gulp-mocha');
 var { restore, build, test, pack, publish } = require('gulp-dotnet-cli');
 
+// add .bin to PATH
+function getPathVariableName() {
+    // windows calls it's path 'Path' usually, but this is not guaranteed.
+    if (process.platform === 'win32') {
```
dsgouda: Can you explain what's happening here?

olydis: this function gets the PATH variable for the respective OS.
While it's (seemingly) always `PATH` on Linux, on Windows the capitalization can differ!
So we search for any variable that matches "path" (case insensitively) or fall back to `Path` which seems to be the usual casing in Win10

olydis:
@dsgouda Thanks!

PR #100, opened 2017-09-19
Add new linter rules for detecting various types of documentation errors.

sptramer:
This is work that came out of the OneWeek proof-of-concept. There will be additional rules added, but this is the starting point. This pull request:

* Adds a new type of API check, `.doc`
* Allows for rules to have an optional `cleanup()` method that is called when done processing a document
* Adds the `DocViolation` violation type
* Introduces new rules: D4002, D4003, D4004, D4100. A previous rule regarding descriptions (`DescriptionMustNotBeNodeName`) has been changed to D4001 as it is a documentation rule, not an API violation.
* D4100 runs an external program, `proselint`, which can take a fair amount of time. This rule will need some adjustments to increase efficiency (such as validating an entire set of `description` values all at once, rather than spawning a process individually.)
* Tests for each new rule.

There is some additional work that will be in a future PR that will need to be discussed with the team:

* The addition of an `x-ms-disable-lint: string[]` extension in the Swagger, which will be a list of rules by identifier code to explicitly disable. We will want to discuss this in depth to ensure that there is no abuse.
* Splitting the documentation rules into two tiers, one which is a set of comparatively simpler checks that we know need to be run with every checkin, and ones that should only be nightly.
* More rules!


sptramer:
To the reviewer: Please let me know if I need to set up additional CI testing for proselint vs. non-proselint systems. Currently the test passes on both, but does not distinguish between behavior on these systems.

src/typescript/azure-openapi-validator/index.ts
```diff
@@ -7,18 +7,23 @@ import { Message } from "../jsonrpc/types";
 import { rules, OpenApiTypes, MergeStates } from "./rule";
 
 // register rules
-require("./rules/DescriptionMustNotBeNodeName");
-require("./rules/ControlCharactersAreNotAllowed");
-require("./rules/ArraySchemaMustHaveItems");
-require("./rules/PostOperationIdContainsUrlVerb");
-require("./rules/LicenseHeaderMustNotBeSpecified");
+let fs = require('fs');
```
fearthecowboy: direct use of `require` in TypeScript is discouraged.  
```  typescript
let fs = require('fs');
let path = require('path');
```

should be at *least* referenced as follows:
``` typescript
import * as fs from 'fs'
import * as path from 'path'
```

And more preferably, just pulling in the functions that are used:
``` typescript
import { join } from 'path'
```

src/typescript/azure-openapi-validator/index.ts
```diff
@@ -7,18 +7,23 @@ import { Message } from "../jsonrpc/types";
...
+let path = require('path');
+
+for (const ruleFile of fs.readdirSync(path.join(__dirname, "rules"))) {
```
fearthecowboy: I'm not a big fan of implicit discovery -- and this causes trouble with some work I'm doing with building a static-linker for nodejs.

As much as I can appreciate the desire to not have to have an explicit reference to the individual rule files, I'd prefer that over using `readdirSync` and arbitrarily using `require` on the discovered files.

sptramer: The problem with non-implicit loading in this case is that there are up to 28(!) additional rules requested by various teams (if none of them are combined, and are all viable automated checks.) These could be placed in a single file for re-export to the `run` method, but the problem there comes from two sources:

1. Node's resolution at runtime problem (no static analysis to see if a referenced import exists)
2. The extremely large number of manually managed imports would make it easy to lose track of them.

I'm curious about your thoughts given these constraints and challenges. If we just need to suck it up and manually manage everything because that's the required style for the project, that's fine.

fearthecowboy: Perhaps the halfway point would be a build-time script that discovers them, and generates a single file that `require`s them all. Then just require that one file and be done with it.

That way, you can make it not a manual process, and we can still target the validator for static-linking.

sptramer: Good suggestion - I'll make this change.

gulpfile.js
```diff
@@ -26,11 +27,28 @@ process.env[getPathVariableName()] = path.join(__dirname, "src/dotnet/AutoRest/n
 // All the typescript tasks
 gulp.task('clean/typescript', function () {
     console.log('Cleaning build directories...');
-    return gulp.src(['src/typescript/azure-openapi-validator/*.{js,d.ts}', 'src/typescript/jsonrpc/*.{js,d.ts}'], { read: false })
+    return gulp.src(['src/typescript/**/*.{js,d.ts}'], { read: false })
         .pipe(clean({ force: true }));
 });
 
-gulp.task('build/typescript', ['clean/typescript'], function () {
+gulp.task('generate/typescript', ['clean/typescript'], function() {
```
fearthecowboy: :+1:
LGTM

veronicagg:
@dsgouda do you mind taking a look and providing some feedback on this PR?

src/typescript/azure-openapi-validator/rule.ts
```diff
@@ -7,7 +7,8 @@ import { JsonPath } from "../jsonrpc/types";
 export enum OpenApiTypes {
   "default" = 1 << 0,
   "arm" = 1 << 1,
-  "dataplane" = 1 << 2
+  "dataplane" = 1 << 2,
+  "doc" = 1 << 3,
```
veronicagg: why is doc an OpenApiType? The intent of this enum is to indicate what type of spec we're dealing with. So far we know of just a basic regular OpenAPI doc, let's say not related to Azure, and Azure related ones, into to categories, arm and dataplane. Doc doesn't seem to map to these categories.

src/typescript/azure-openapi-validator/rules/DescriptionAvoidsGerunds.ts
```diff
@@ -0,0 +1,36 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+import { MergeStates, OpenApiTypes, rules } from '../rule';
+
+export const DescriptionAvoidsGerunds: string = "DescriptionAvoidsGerunds";
```
veronicagg: have you run this rule on the specs in the repo? how many warnings are getting flagged?
We don't have a way currently to filter per category so, I worry about the noise brought by these, if we don't introduce something like that.

src/typescript/azure-openapi-validator/rules/DescriptionMustNotBeNodeName.ts
```diff
@@ -3,41 +3,40 @@
...
 rules.push({
-  id: "R3011",
+  id: "D4001",
   name: DescriptionMustNotBeNodeName,
   severity: "error",
-  category: "RPCViolation",
+  category: "DocViolation",
   mergeState: MergeStates.composed,
-  openapiType: OpenApiTypes.arm,
+  openapiType: OpenApiTypes.doc,
```
veronicagg: does this apply to ARM and data plane specs, any spec? if so, the value here may need to be default or "arm and dataplane"

src/typescript/azure-openapi-validator/.gitignore
```diff
@@ -0,0 +1,3 @@
+# Ignore autogenerated files
+
+rules/index.ts
```
dsgouda: Why do we need this again?

src/typescript/azure-openapi-validator/rule.ts
dsgouda: Don't see the harm in being descriptive here, i.e., `documentation`

src/typescript/azure-openapi-validator/rule.ts
```diff
@@ -23,14 +24,16 @@ interface ValidationMessage {
 export interface Rule {
   readonly id: string; // see Rxxx/Sxxx codes on https://github.com/Azure/azure-rest-api-specs/blob/master/documentation/openapi-authoring-automated-guidelines.md
   readonly name: string; // see same website as above
-  readonly category: ("RPCViolation" | "OneAPIViolation" | "SDKViolation");
+  readonly category: ("RPCViolation" | "OneAPIViolation" | "SDKViolation" | "DocViolation");
```
dsgouda: Same, `DocumentationViolation`

src/typescript/azure-openapi-validator/rules/DescriptionAvoidsGerunds.ts
```diff
@@ -0,0 +1,36 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+import { MergeStates, OpenApiTypes, rules } from '../rule';
```
dsgouda: Yay for typescript rules!

src/typescript/azure-openapi-validator/rules/DescriptionAvoidsGerunds.ts
```diff
@@ -0,0 +1,36 @@
...
+    if (typeof (node) !== 'string') {
+      return;
+    }
+
+    const nodeValue: string = <string>node;
+
+    // Known good words which are not gerunds, but do end in -ing
+    const goodWords = ["string"];
+
+    var gerunds = nodeValue.match(/\b\w+ing\b/g);
```
dsgouda: Are we assuming that description will be exactly one word here?

src/typescript/azure-openapi-validator/rules/DescriptionAvoidsGerunds.ts
```diff
@@ -0,0 +1,36 @@
...
+    if (gerunds != null) {
+      let filtered = gerunds.filter(word => goodWords.indexOf(word.toLowerCase()) == -1);
+      if (filtered.length > 0) {
+        yield { message: `${msg} -- Possible gerunds are ${filtered}`, location: path };
```
dsgouda: nit: Better description:
`${msg} -- Gerunds detected: ${filtered}`

src/typescript/azure-openapi-validator/rules/DescriptionAvoidsGerunds.ts
```diff
@@ -0,0 +1,36 @@
...
+      return;
+    }
+
+    const nodeValue: string = <string>node;
+
+    // Known good words which are not gerunds, but do end in -ing
+    const goodWords = ["string"];
+
+    var gerunds = nodeValue.match(/\b\w+ing\b/g);
+    if (gerunds != null) {
```
dsgouda: strict checking is recommended
`!==`

src/typescript/azure-openapi-validator/rules/DescriptionAvoidsGerunds.ts
veronicagg: could you include comments on intent of the rules for each of them? It helps understand what the rule is supposed to do from start, even when the name is not as obvious.

src/typescript/azure-openapi-validator/rules/DescriptionNoIdenticalSiblings.ts
```diff
@@ -0,0 +1,68 @@
...
+
+export const DescriptionNoIdenticalSiblings: string = "DescriptionNoIdenticalSiblings";
+const minimumDescriptionLength = 20;
+
+var checkedPaths: string[] = [];
+
+rules.push({
+  id: "D4003",
+  name: DescriptionNoIdenticalSiblings,
+  severity: "error",
```
veronicagg: do you know how many of these errors are being currently detected in the repo?

src/typescript/azure-openapi-validator/rules/DescriptionTooShort.ts
```diff
@@ -0,0 +1,33 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { MergeStates, OpenApiTypes, rules } from '../rule';
+import { trimDescription } from './utilities';
+
+export const DescriptionTooShort: string = "DescriptionTooShort";
+const minimumDescriptionLength = 20;
```
veronicagg: where does this number come from?
Is this rule accurate enough to be an error?

src/typescript/azure-openapi-validator/rules/LintDescriptionProse.ts
```diff
@@ -0,0 +1,53 @@
...
+rules.push({
+  id: "D4100",
+  name: LintDescriptionProse,
+  severity: "warning",
+  category: "DocViolation",
+  mergeState: MergeStates.composed,
+  openapiType: OpenApiTypes.doc,
+  appliesTo_JsonQuery: "$..description",
+  run: function* (doc, node, path) {
+    const msg: string = "`proselint` generated warnings: ";
```
veronicagg: what type of things it proselint checking? do you have a link to extra documentation on it?

src/typescript/azure-openapi-validator/rules/DescriptionNoIdenticalSiblings.ts
```diff
@@ -0,0 +1,68 @@
...
+  category: "DocViolation",
+  mergeState: MergeStates.composed,
+  openapiType: OpenApiTypes.doc,
+  appliesTo_JsonQuery: "$..*[?(@.description)]",
```
dsgouda: Maybe we can modify this query to pick all the parent nodes to begin with instead of going to each `description` node and then trying to validate its parent node

src/typescript/azure-openapi-validator/rules/DescriptionTooShort.ts
```diff
@@ -0,0 +1,33 @@
...
+
+rules.push({
+  id: "D4002",
+  name: DescriptionTooShort,
+  severity: "error",
```
dsgouda: Do we really want this to be an error?

src/typescript/azure-openapi-validator/rules/DescriptionTooShort.ts
```diff
@@ -0,0 +1,33 @@
...
+  appliesTo_JsonQuery: "$..description",
+  run: function* (doc, node, path) {
+    const msg: string = "Description length is too short.";
+    if (typeof (node) !== 'string') {
+      return;
+    }
+
+    // Skip 0-length nodes; these are picked up by another validator.
+    const description: string = trimDescription(node);
+    if (node.length > 0 && description.length < minimumDescriptionLength) {
```
dsgouda: Maybe check description for null too

src/typescript/azure-openapi-validator/rules/LintDescriptionProse.ts
```diff
@@ -0,0 +1,53 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+import { MergeStates, OpenApiTypes, rules } from '../rule';
+export const LintDescriptionProse: string = "LintDescriptionProse";
```
veronicagg: this rules appears that could expose a lot of violations, I worry about noise, not having a way to filter doc violations.

src/typescript/azure-openapi-validator/tests/individual-azure-tests.ts
```diff
@@ -28,10 +28,11 @@ import { LicenseHeaderMustNotBeSpecified } from '../rules/LicenseHeaderMustNotBe
     const messages: Message[] = await collectTestMessagesFromValidator(fileName, OpenApiTypes.arm, MergeStates.individual);
```
dsgouda: nit: No useful changes

src/typescript/azure-openapi-validator/tests/composite-azure-test.ts
```diff
@@ -1,22 +0,0 @@
-/*---------------------------------------------------------------------------------------------
- *  Copyright (c) Microsoft Corporation. All rights reserved.
```
dsgouda: Maybe not delete this file altogether. Can we at least add a dummy test case so it becomes a placeholder for any new composite tests we would like to add later

src/typescript/azure-openapi-validator/tests/default-tests.ts
```diff
@@ -19,4 +19,4 @@ import { ArraySchemaMustHaveItems } from '../rules/ArraySchemaMustHaveItems';
     const messages: Message[] = await collectTestMessagesFromValidator(fileName, OpenApiTypes.default, MergeStates.composed);
```
dsgouda: nit: no useful changes

veronicagg:
Few more comments: 
- The following doc would need to be updated accompanying these rules https://github.com/Azure/azure-rest-api-specs/blob/current/documentation/openapi-authoring-automated-guidelines.md
- We may want to not enable the proselint rule by default and have some way to explicitly invoke it. 
- regarding your question on disabling rules, we have a mechanism described at https://github.com/Azure/autorest/tree/97b68250afd96111f79047e24e22eeb82a30426f/src/autorest-core/test/variations/suppressions#suppressions , would this work for you?
- Re: splitting doc rules in simpler checks versus more complicated ones, yes, or you may want to create a specific doc validator plugin to autorest with its own settings, so it can be invoked as desired and controlled by the configuration in readme.md file for each case. Have you evaluated this? https://github.com/Azure/autorest/blob/f547506b802a19e2ab6ce24f539b72bf7519267c/docs/developer/guide/writing-a-new-plugin.md

@salameer @kirthik doc rules coming in, what should we be aware in terms of introducing these? I've added questions on the PR already around my concerns on noise, feel free to provide your comments.

src/typescript/azure-openapi-validator/rule.ts
dsgouda: I missed this. We do not want to introduce a new OpenAPI type here, only a new category for the rules themselves.

sptramer: Will be removed. Since these rules apply to all OpenAPI types, I'm assuming that the rules will need to be `|`ed together, rather than set to `default`?

src/typescript/azure-openapi-validator/rules/DescriptionAvoidsGerunds.ts
sptramer: There hasn't been a test run on the full swagger yet. While all of these rules will undoubtedly flag a bunch of stuff, they have been required for that exact reason. Running on live data may be required for gerund detection, so that I can whitelist any other non-gerund `-ing` nouns.

src/typescript/azure-openapi-validator/rules/DescriptionTooShort.ts
sptramer: The length was requested to be an error by PMs. Too-short descriptions mean a lack of contextual information.

The specific length in this case was driven by a thought experiment of "what would be the shortest possible description, including abbreviations and acronyms, which would be useful." Production data was not used under the assumption that it would take a long time to comb through.

src/typescript/azure-openapi-validator/rules/LintDescriptionProse.ts
sptramer: The full `proselint` documentation is available. https://github.com/amperser/proselint/

src/typescript/azure-openapi-validator/rules/LintDescriptionProse.ts
sptramer: Please see the comment associated with the PR. When additional rules are added, I would like to spec in `x-ms-disable-lint` as a Swagger extension which lists rules to ignore by ID.

src/typescript/azure-openapi-validator/.gitignore
sptramer: This avoids checking in the auto-generated rules indexing file produced by `gulp generate/typescript`.

src/typescript/azure-openapi-validator/rules/DescriptionAvoidsGerunds.ts
sptramer: No; this matches individual words within the string (`\b` indicates a word boundary.) While this will also detect terms other than gerunds such as present participle verbs, it's unfortunately the best check that can be done without extensive syntax analysis (which is an option that may be added if an excessive number of present participles are detected - though these are ALSO very bad for machine translation.)

src/typescript/azure-openapi-validator/rules/DescriptionAvoidsGerunds.ts
sptramer: Factually inaccurate. `-ing` terms may also be present participles or other elements of English grammar, which the documentation writer will need to manually check. See above regarding more advanced sentence diagramming or NLP tasks which might be able to distinguish between them.

src/typescript/azure-openapi-validator/rules/DescriptionNoIdenticalSiblings.ts
sptramer: As stated in the code itself:

```
// JSONPath does not have a parent operator, so we need to do some manipulation
// of the 'path', along with an additional search (and keep a record of the parents)
// in order to compare multiple nodes at once.
```

The primary issue is that JSONPath will not accept `@.*.description` as a query, when this is the exact term required here - we care about any of the current node's children having a description.

src/typescript/azure-openapi-validator/rules/DescriptionTooShort.ts
sptramer: See above regarding my comment about short descriptions often lacking contextual information. I'd be willing to introduce a new `variable` severity type (where the exact severity is returned in the yielded response.) Then abnormally short descriptions could be errors, and descriptions under a "reasonable" length would be warnings. Future proposed rules will also require variable severity.

src/typescript/azure-openapi-validator/tests/individual-azure-tests.ts
sptramer: Moved an existing test to its new designation as a documentation violation (the "no description" test). If you would rather keep the original designation, then I'll move it back here.

src/typescript/azure-openapi-validator/tests/composite-azure-test.ts
sptramer: Is there a reason to have this file around other than as a dummy?

sptramer:
@veronicagg -

* I'll update the authoring guidelines when this PR is accepted, so that it can be merged in concert.
* `proselint` is definitely what the DevEx team has been referring to as a "Level 2" rule for the ref, which is a rule intended to be run only nightly. Originally I was going to control this by adding in an additional `OpenApiType` enum, but am now willing to take suggestions on how to segment these rules into two categories. Is writing a second plugin the ideal way to do this?
* `suppress` should work! It will need to be able to be applied to individual nodes, though - it looks like it is either a top-level directive, or takes a JsonPath?

src/typescript/azure-openapi-validator/rule.ts
dsgouda: That depends on the rule. Given you are dealing with descriptions here I'd say its got to do with arm openapitype, however, confirm this with @kirthik

src/typescript/azure-openapi-validator/rules/DescriptionAvoidsGerunds.ts
dsgouda: I'm happy with the regex.

src/typescript/azure-openapi-validator/rules/DescriptionNoIdenticalSiblings.ts
dsgouda: D'oh! did not know that!

src/typescript/azure-openapi-validator/rules/DescriptionTooShort.ts
dsgouda: Also, shouldn't you trim this?

src/typescript/azure-openapi-validator/tests/composite-azure-test.ts
dsgouda: Not that I can think of.

sptramer: In that case I guess it comes down to what your team's policies are - if you'll accept a dummy file here, I'll totally add it back in.

src/typescript/azure-openapi-validator/rules/DescriptionTooShort.ts
sptramer: Trimmed in the line above, but that's done before any nil check. Will do both here.

sptramer: Actually just did a full context check here - neither a nil check or a trim are necessary. The `typeof (node)` is checked earlier for `string`, so it must be non-null.

src/typescript/azure-openapi-validator/rules/DescriptionAvoidsGerunds.ts
sptramer: Fixed

src/typescript/azure-openapi-validator/rule.ts
sptramer: Fixed

src/typescript/azure-openapi-validator/rules/DescriptionAvoidsGerunds.ts
sptramer: Fixed

src/typescript/azure-openapi-validator/rules/DescriptionTooShort.ts
dsgouda: Please confirm with @kirthik

src/typescript/azure-openapi-validator/rules/DescriptionNoIdenticalSiblings.ts
sptramer: It totally didn't make sense to me either, and sucks. Especially since other JSON search specs do include the ability to query on available children/get a parent.

src/typescript/azure-openapi-validator/tests/composite-azure-test.ts
dsgouda: Delete this for now, guess we can check it from this commit letter.

src/typescript/azure-openapi-validator/rules/DescriptionNoIdenticalSiblings.ts
dsgouda: I hear ya

sptramer:
@dsgouda @verionicagg - Ready for re-review - going to start a run of the new rules on the entire codebase (minus proselint, for speed) in the meantime.

dsgouda:
@sptramer I believe you are still in a conversation with @kirthik regarding whether to introduce these rules and with what severity. Please get a confirmation from here before we can proceed.

veronicagg:
@sptramer let's sync up in a meeting to agree on next steps, I'll set something up.

sptramer:
I was able to run the rules on the entire `azure-rest-api-spec` repo, and these are the final results:
---

```
Time measured the batch script at 7:03 real-time. Not bad for processing this amount of data.
./lintify.sh  209.14s user 105.16s system 74% cpu 7:03.61 total

Surfaces approximately 9911 errors ("DocumentationViolation")

D401 (Not node name): 64
D402 (Short description): 5735
D403 (Identical siblings): 302
D404 (Gerunds): 3810

SHORT DESCRIPTION HISTOGRAM
----
       # Length
    992 2
     35 3
     15 4
     16 5
     14 6
    105 7
    278 8
    210 9
    361 10
    666 11
    196 12
    478 13
    227 14
    416 15
    323 16
    421 17
    775 18
    590 19
```

sptramer:
Closing this PR in favor of creating a new validation tool which will be specifically for documentation.

fearthecowboy:
@sptramer  that's probably a great idea.

Issue #101, opened 2017-09-27
Fix path reported by PostOperationIdContainsUrlVerb


dsgouda:
Moving from https://github.com/Azure/autorest/issues/2617

PR #103, opened 2017-09-27
Returning correct path from PostOperationIdContainsUrlVerb

dsgouda:
Addressing https://github.com/Azure/azure-openapi-validator/issues/101


bgold09:
Does this also fix the issue where the verb the rules suggests is actually the path to the operation in the spec? It is not obvious how this fixes the issue. 

Should a new test be added to prevent a regression, or do existing tests need to be fixed?

dsgouda:
@bgold09 Yes this fixes that exact issue, the rule was trying to find a node with the name of the operationId instead of `operationId` itself.
I don't think we have tests to detect correctness of paths (which seems like an overkill) and we are thinking of demoting issues raised by paths not found [here](https://github.com/Azure/autorest/pull/2590) anyway

PR #108, opened 2017-10-19
Fix errors in the application of the 'NotNodeName' rule.

sptramer:
Changes ported over from `new-linter-rules` that affect an existing rule's ability to correctly detect documentation problems.


sptramer:
Yes, please merge!

PR #112, opened 2017-10-27
Add path parameters should not be empty rule

vladbarosan:


vladbarosan:
There was some github page I think where all errors were documented. We do we need to update that one ?

dsgouda:
Documentation needs to go [here](https://github.com/Azure/azure-rest-api-specs/blob/current/documentation/openapi-authoring-automated-guidelines.md)

vladbarosan:
This addresses issue Azure/azure-rest-api-specs#1918

src/typescript/azure-openapi-validator/rules/PathParameterMustNotBeEmpty.ts
```diff
@@ -0,0 +1,25 @@
...
+import { MergeStates, OpenApiTypes, rules } from '../rule';
+export const PathParametersMustNotBeEmpty: string = "PathParametersMustNotBeEmpty";
+
+rules.push({
+  id: "R2067",
+  name: PathParametersMustNotBeEmpty,
+  severity: "error",
+  category: "RPCViolation",
+  mergeState: MergeStates.individual,
+  openapiType: OpenApiTypes.arm,
```
dsgouda: You may want to check if it makes sense to extend it to Dataplane

src/typescript/azure-openapi-validator/rules/PathParameterMustNotBeEmpty.ts
```diff
@@ -0,0 +1,25 @@
...
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+import { MergeStates, OpenApiTypes, rules } from '../rule';
+export const PathParametersMustNotBeEmpty: string = "PathParametersMustNotBeEmpty";
+
+rules.push({
+  id: "R2067",
+  name: PathParametersMustNotBeEmpty,
+  severity: "error",
```
dsgouda: I am not convinced that we should flag minLength missing for parameters as an error as of yet, might light up the validation charts wildly, please verify with kirthi

src/typescript/azure-openapi-validator/rules/PathParameterMustNotBeEmpty.ts
```diff
@@ -0,0 +1,25 @@
...
+  category: "RPCViolation",
+  mergeState: MergeStates.individual,
+  openapiType: OpenApiTypes.arm,
+
+  appliesTo_JsonQuery: "$.paths..parameters[?(@.in=='path')]",
+  run: function* (doc, node, path) {
+
+    const pathNodes: string[] = (<string>path[1]).split('/');
```
dsgouda: nit: I don't remember how this works but I remember putting [this](https://github.com/vladbarosan/azure-openapi-validator/blob/0b78dff4cbd2cb851bcf9e445d8f30939f2ab987/src/typescript/azure-openapi-validator/rules/PostOperationIdContainsUrlVerb.ts#L17) comment , please make sure `path[1]` is the right thing to do

src/typescript/azure-openapi-validator/rules/PathParameterMustNotBeEmpty.ts
```diff
@@ -0,0 +1,25 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+import { MergeStates, OpenApiTypes, rules } from '../rule';
+export const PathParametersMustNotBeEmpty: string = "PathParametersMustNotBeEmpty";
+
+rules.push({
+  id: "R2067",
```
dsgouda: nit: make sure you're picking a unique rule id

dsgouda:
Perhaps beyond the purview of your task, but we don't have positive tests for the typescript rules today, if you can implement it, that'd be awesome! You can start with a positive test for just this rule for now.

src/typescript/azure-openapi-validator/rules/PathParameterMustNotBeEmpty.ts
vladbarosan: Yes it is correct. the actual path is always the second element. In that case you referenced it works because it only has 2 elements so the last element is actually path[1].

src/typescript/azure-openapi-validator/rules/PathParameterMustNotBeEmpty.ts
vladbarosan: yes I searched in the source code it was unique. Also checked in the doc link you provided now and it is not referenced anywhere.

src/typescript/azure-openapi-validator/rules/PathParameterMustNotBeEmpty.ts
vladbarosan: It is an error, but I see your point as this will probably blow up the error count from what I saw in the specs. Either way, I will work to fix the specs. So I would say I'll move it to a warning until we get those fixed and then we can change it to an error

dsgouda: Even then, better discuss this with @veronicagg and @kirthik before releasing the rule.

src/typescript/azure-openapi-validator/rules/PathParameterMustNotBeEmpty.ts
vladbarosan: Initial  concern was for Management, but yes I don't see why wouldn't be for Dataplane. Will update

vladbarosan:
Added issue #113 for adding positive test cases.

src/typescript/azure-openapi-validator/rules/PathParameterMustNotBeEmpty.ts
olydis: fun fact: variable not used (and I wouldn't know why) 😉

src/typescript/azure-openapi-validator/rules/PathParameterMustNotBeEmpty.ts
```diff
@@ -0,0 +1,25 @@
...
+  category: "RPCViolation",
+  mergeState: MergeStates.individual,
+  openapiType: OpenApiTypes.arm,
+
+  appliesTo_JsonQuery: "$.paths..parameters[?(@.in=='path')]",
+  run: function* (doc, node, path) {
+
+    const pathNodes: string[] = (<string>path[1]).split('/');
+
+    if (!Object.keys(node).includes("minLength") || node.minLength < 1) {
```
olydis: JavaScript magic: this is equivalent to the (smaller) `!("minLength" in node)` or maybe even better:
``` JavaScript
typeof node.minLength !== "number"
```
That last thing checks the same as you do, but also ensures the type, so the comparison you make afterwards will not be between an apple and a number

src/typescript/azure-openapi-validator/rules/PathParameterMustNotBeEmpty.ts
vladbarosan: yup don't needed it anymore

src/typescript/azure-openapi-validator/rules/PathParameterMustNotBeEmpty.ts
```diff
@@ -0,0 +1,22 @@
...
+  name: PathParametersMustNotBeEmpty,
+  severity: "warning",
+  category: "RPCViolation",
+  mergeState: MergeStates.individual,
+  openapiType: OpenApiTypes.arm,
+
+  appliesTo_JsonQuery: "$.paths..parameters[?(@.in=='path')]",
+  run: function* (doc, node, path) {
+    if ((typeof node.minLength !== "number") || (node.minLength < 1)) {
+      yield { message: `Path parameters should should have the 'minLength' attribute with a value > 0: Parameter '${node.name}' in:'${path[1]}'. `, location: path };
```
veronicagg: Since this is a warning and we're not sure if the service may for some reason want to deal with empty strings, we may want to update the message to something like: "If your service requires non-empty path parameters, please add 'minLength' ... ".
You should have a corresponding PR to update documentation at https://github.com/Azure/azure-rest-api-specs/blob/current/documentation/openapi-authoring-automated-guidelines.md (that you can get reviewed and would merge once this one is merged) - You can explain on the "why" that including this property helps the SDK check for this instead of making a call to the service with an empty string.

konrad-jamrozik:
Closing this PR as ancient.

Issue #106, opened 2017-10-05
GetCamelCasedSuggestion returns erraneous suggestions


dsgouda:
Looks like there was a regression in our latest commit [here](https://github.com/Azure/azure-openapi-validator/commit/73b23ec6c99e5e038816f4bd2e0d01256fd27d6f#diff-ca0f26932fbcc64f5f8b35b071880575) which causes the camel cased suggestion to ignore the first capitalized letter.
Eg: Camelcased suggestion for `ConsoleApplication` should be `consoleApplication` but the current implementation returns `ConsoleApplication`

We `must` implement more test cases like [these](https://github.com/Azure/azure-openapi-validator/blob/73b23ec6c99e5e038816f4bd2e0d01256fd27d6f/src/dotnet/OpenAPI.Validator.Tests/GetCamelCasedSuggestionTests.cs) for camelcases at least

PR #131, opened 2017-11-24
Suggest correct camel case

mcardosos:
Fixes https://github.com/Azure/azure-openapi-validator/issues/106
PTAL @veronicagg

### DIFF
--- a/src/dotnet/OpenAPI.Validator/Model/Utilities/ValidationUtilities.cs
+++ b/src/dotnet/OpenAPI.Validator/Model/Utilities/ValidationUtilities.cs
@@ -27,7 +27,7 @@ namespace OpenAPI.Validator.Model.Utilities
         public static IEnumerable<string> GetTenantResourceModels(IEnumerable<string> resourceModels, ServiceDefinition serviceDefinition)
         {
             List<string> tenantResourceModels = new List<string>();
-            if(serviceDefinition.Paths.Any())
+            if (serviceDefinition.Paths.Any())
             {
                 foreach (KeyValuePair<string, Dictionary<string, Operation>> path in serviceDefinition.Paths)
                 {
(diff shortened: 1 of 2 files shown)


src/dotnet/OpenAPI.Validator/Model/Utilities/ValidationUtilities.cs
```diff
@@ -375,9 +375,11 @@ public static string GetCamelCasedSuggestion(string name)
             => Regex.Replace(
                 name,
                 @"(?x)
-                    (?<=\p{Lu}) # after any uppercase character,
+                    ((?<=\p{Lu}) # after any uppercase character,
                     \p{Lu}+     # find a series of uppercase characters,
-                    (?=\p{Lu}|\z) # followed by an uppercase character or the end of the string
+                    (?=\p{Lu}|\z)) # followed by an uppercase character or the end of the string
+                    |(\A[A-Z]) # or it happens to be the first character
+                    |((?<=[a-z]){1}[A-Z]\z) # or it is an uppercase at the end, next to a lowercase
```
veronicagg: which case is this last one for? example?

src/dotnet/OpenAPI.Validator/Model/Utilities/ValidationUtilities.cs
```diff
@@ -375,9 +375,11 @@ public static string GetCamelCasedSuggestion(string name)
             => Regex.Replace(
```
veronicagg: It looks like we need to update the comments for this method, to include the updated examples.

src/dotnet/OpenAPI.Validator/Model/Utilities/ValidationUtilities.cs
mcardosos: `ManyThingS` becomes `manyThings`

src/dotnet/OpenAPI.Validator/Model/Utilities/ValidationUtilities.cs
mcardosos: You mean the ones after each regex section?

veronicagg: lines https://github.com/Azure/azure-openapi-validator/pull/131/files/d018e7f04bb49b7add7b373f35eb15ad1f722fd9#diff-ca0f26932fbcc64f5f8b35b071880575R369 and https://github.com/Azure/azure-openapi-validator/pull/131/files/d018e7f04bb49b7add7b373f35eb15ad1f722fd9#diff-ca0f26932fbcc64f5f8b35b071880575R370

src/dotnet/OpenAPI.Validator/Model/Utilities/ValidationUtilities.cs
veronicagg: I see, ok, honestly we could also keep it as it, but fine, it's a suggestion, so the user evaluates at the end anyway :)

src/dotnet/OpenAPI.Validator/Model/Utilities/ValidationUtilities.cs
mcardosos: FIxed

PR #161, opened 2018-04-27
Add pageable rule

mcardosos:
Big question, what is the logic behind the linter rules IDs?

### DIFF
--- a/docs/pageable-operations.md
+++ b/docs/pageable-operations.md
@@ -1,0 +1,134 @@
+# PageableOperations
+
+## Description
+
+This rule appears when you define a get operation, and its response schema has a property that is an array. The operation might be pageable.
+
+## How to fix
+
+Add an `x-ms-pageable` extension to the operation.
+
+## Effect on generated code
+
+### Before
+
+#### Spec
+
+```json5
+…
+    "paths": {
+        "/plants": {
+            "get": {
+                "operationId": "Stuff_List",
+                "responses": {
+                    "200": {
+                        "schema": {
+                            "$ref": "#/definitions/StuffList"
+                        }
+                    }
+                }
+            }
+        }
+    }
+…
+    "definitions": {
+        "StuffList": {
+            "type": "object",
+            "properties": {
+                "value": {
+                    "type": "array",
+                    "items": {
+                        "$ref": "#/definitions/Stuff"
+                    }
+                },
+                "nextLink": {
+                    "type": "string",
+                }
+            }
+        },
+        "Stuff": {
+            "type": "object",
+            "properties": {
+                "name": {
+                    "type": "string",
+                },
+            }
+        }
+    }
+…
+```
+
+#### Generated code
+
+```csharp
+public static ProductResult GetMultiplePages(this IPagingOperations operations, string clientRequestId = default(string), PagingGetMultiplePagesOptions pagingGetMultiplePagesOptions = default(PagingGetMultiplePagesOptions))
+{
+    return operations.GetMultiplePagesAsync(clientRequestId, pagingGetMultiplePagesOptions).GetAwaiter().GetResult();
+}
+```
+
+### After
+
+#### Spec
+
+```json5
+…
+    "paths": {
+        "/plants": {
+            "get": {
+                "operationId": "Stuff_List",
+                "responses": {
+                    "200": {
+                        "schema": {
+                            "$ref": "#/definitions/StuffList"
+                        }
+                    }
+                }
+                "x-ms-pageable": {
+                    "nextLinkName": "nextLink"
+                }
+            }
+        }
+    }
+…
+    "definitions": {
+        "StuffList": {
+            "type": "object",
+            "properties": {
+                "value": {
+                    "type": "array",
+                    "items": {
+                        "$ref": "#/definitions/Stuff"
+                    }
+                },
+                "nextLink": {
+                    "type": "string",
+                }
+            }
+        },
+        "Stuff": {
+            "type": "object",
+            "properties": {
+                "name": {
+                    "type": "string",
+                },
+            }
+        }
+    }
+…
+```
+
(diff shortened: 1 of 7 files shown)


src/typescript/azure-openapi-validator/tests/composite-azure-test.ts
```diff
@@ -12,11 +12,18 @@ import {
...
+import { PageableOperation } from '../rules/PageableOperation';
 
 @suite class CompositeAzureTests {
   @test async "description should not be property name"() {
     const fileName: string = 'DescriptionSameAsPropertyName.json';
     const messages: Message[] = await collectTestMessagesFromValidator(fileName, OpenApiTypes.arm, MergeStates.composed);
     assertValidationRuleCount(messages, DescriptionMustNotBeNodeName, 2);
   }
+
+  @test async "operations returning a model including an array might be pageable"() {
```
sarangan12: Can you add a positive test case too?

src/typescript/azure-openapi-validator/rules/PageableOperation.ts
```diff
@@ -0,0 +1,39 @@
...
+  appliesTo_JsonQuery: "$.paths[?(@.get)]",
+  run: function* (doc, node, path) {
+    const operations = Object.keys(node);
+    const getKey = operations.find(key => {
+      return key.toLowerCase() === 'get';
+    });
+
+    const schemaRef = node[getKey]['responses']['200']['schema']['$ref'];
+    const schemaPath: string[] = (<string>schemaRef).split('/');
+    const schemaProperties = doc.definitions[schemaPath[2]].properties;
```
sarangan12: What will happen if the model is defined inline?

src/typescript/azure-openapi-validator/rules/PageableOperation.ts
```diff
@@ -0,0 +1,39 @@
...
+
+    const valueKey = Object.keys(schemaProperties).find(key => {
+      return key.toLowerCase() === 'value';
```
sarangan12: Is the key - 'value' a mandatory field? Please confirm if an array could be defined without this key

src/typescript/azure-openapi-validator/rules/PageableOperation.ts
```diff
@@ -0,0 +1,39 @@
...
+    });
+
+    if (valueKey != undefined) {
```
sarangan12: Will this be true if the valueKey is null? If so, I am fine

docs/pageable-operations.md
```diff
@@ -0,0 +1,134 @@
...
+            "properties": {
+                "value": {
+                    "type": "array",
+                    "items": {
+                        "$ref": "#/definitions/Stuff"
+                    }
+                },
+                "nextLink": {
+                    "type": "string",
+                }
```
dsgouda: nit: Let's be diligent, mark this as `readonly`

src/typescript/azure-openapi-validator/rules/PageableOperation.ts
```diff
@@ -0,0 +1,39 @@
...
+  mergeState: MergeStates.composed,
+  openapiType: OpenApiTypes.arm | OpenApiTypes.dataplane,
+  appliesTo_JsonQuery: "$.paths[?(@.get)]",
+  run: function* (doc, node, path) {
+    const operations = Object.keys(node);
+    const getKey = operations.find(key => {
+      return key.toLowerCase() === 'get';
+    });
+
+    const schemaRef = node[getKey]['responses']['200']['schema']['$ref'];
```
dsgouda: Most operations should have a `200` response but what happens here if they don't?

src/typescript/azure-openapi-validator/rules/PageableOperation.ts
dsgouda: I think we can run with this assumption since there are linter rules in place to flag inline definitions
On the other hand though, moving this to a utility method could be helpful

src/typescript/azure-openapi-validator/rules/PageableOperation.ts
```diff
@@ -0,0 +1,39 @@
...
+    const schemaPath: string[] = (<string>schemaRef).split('/');
+    const schemaProperties = doc.definitions[schemaPath[2]].properties;
+
+    const valueKey = Object.keys(schemaProperties).find(key => {
+      return key.toLowerCase() === 'value';
+    });
+
+    if (valueKey != undefined) {
+      if (schemaProperties[valueKey].type === 'array') {
+        yield { message: `Based on the response model schema, operation '${node[getKey].operationId}' might be pageable. Consider adding the x-ms-pageable swagger extension.`, location: path.concat(getKey) };
```
dsgouda: nit: let's not call it swagger 😄

src/typescript/azure-openapi-validator/rules/PageableOperation.ts
```diff
@@ -0,0 +1,39 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+import { MergeStates, OpenApiTypes, rules } from '../rule';
```
dsgouda: If I remember correctly, we decided to flag this only if a model had less than 3 properties with one being the value array. There might be models which have a value array and a whole lot of other properties, should we flag this rule for such cases too?

mcardosos: I had not heard the 3 properties thing, but it makes sense

dsgouda: It would be great if you can talk to one of the PMs regarding this before going through with the implementation

PR #179, opened 2018-09-27
Code changes for Issue 172

sarangan12:
Related to issue: https://github.com/Azure/azure-openapi-validator/issues/172

@lmazuel @veronicagg @salameer Please review and approve.

### DIFF
--- a/src/dotnet/OpenAPI.Validator/Properties/Resources.resx
+++ b/src/dotnet/OpenAPI.Validator/Properties/Resources.resx
@@ -361,6 +361,6 @@
     <value>The parameter '{0}' is defined in global parameters section without 'x-ms-parameter-location' extension. This would add the parameter as the client property. Please ensure that this is exactly you want. If so, apply the extension "x-ms-parameter-location": "client". Else, apply the extension "x-ms-parameter-location": "method".</value>
   </data>
   <data name="LongRunningOperationsOptionsValidatorMessage" xml:space="preserve">
-    <value>A LRO Post operation with return value must have "x-ms-long-running-operation-options" extension enabled.</value>
+    <value>A LRO Post operation with return schema must have "x-ms-long-running-operation-options" extension enabled.</value>
   </data>
 </root>
--- a/src/dotnet/OpenAPI.Validator/Properties/Resources.resx
+++ b/src/dotnet/OpenAPI.Validator/Properties/Resources.resx
@@ -360,4 +360,7 @@
   <data name="XmsParameterLocation" xml:space="preserve">
     <value>The parameter '{0}' is defined in global parameters section without 'x-ms-parameter-location' extension. This would add the parameter as the client property. Please ensure that this is exactly you want. If so, apply the extension "x-ms-parameter-location": "client". Else, apply the extension "x-ms-parameter-location": "method".</value>
   </data>
+  <data name="LongRunningOperationsOptionsValidatorMessage" xml:space="preserve">
+    <value>A LRO Post operation with return value must have "x-ms-long-running-operation-options" extension enabled.</value>
+  </data>
 </root>
--- a/src/dotnet/OpenAPI.Validator/Validation/LongRunningOperationsOptionsValidator.cs
+++ b/src/dotnet/OpenAPI.Validator/Validation/LongRunningOperationsOptionsValidator.cs
@@ -1,0 +1,89 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using AutoRest.Core.Logging;
+using OpenAPI.Validator.Model;
+using OpenAPI.Validator.Properties;
+using OpenAPI.Validator.Validation.Core;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace OpenAPI.Validator.Validation
+{
+    public class LongRunningOperationsOptionsValidator : TypedRule<Dictionary<string, Operation>>
+    {
+        /// <summary>
+        /// Id of the Rule.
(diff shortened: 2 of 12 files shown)


src/dotnet/OpenAPI.Validator/Properties/Resources.resx
```diff
@@ -360,4 +360,7 @@
   <data name="XmsParameterLocation" xml:space="preserve">
     <value>The parameter '{0}' is defined in global parameters section without 'x-ms-parameter-location' extension. This would add the parameter as the client property. Please ensure that this is exactly you want. If so, apply the extension "x-ms-parameter-location": "client". Else, apply the extension "x-ms-parameter-location": "method".</value>
   </data>
+  <data name="LongRunningOperationsOptionsValidatorMessage" xml:space="preserve">
+    <value>A LRO Post operation with return value must have "x-ms-long-running-operation-options" extension enabled.</value>
```
veronicagg: should it be "with return schema"? or "type" or "model"?

veronicagg: Please update documentation at https://github.com/Azure/azure-rest-api-specs/blob/master/documentation/openapi-authoring-automated-guidelines.md

veronicagg: could we indicate which LRO operation in the message and maybe which return type? "LRO operation X wich returns model Y must have..."?

src/dotnet/OpenAPI.Validator/Validation/LongRunningOperationsOptionsValidator.cs
```diff
@@ -0,0 +1,89 @@
...
+using AutoRest.Core.Logging;
+using OpenAPI.Validator.Model;
+using OpenAPI.Validator.Properties;
+using OpenAPI.Validator.Validation.Core;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace OpenAPI.Validator.Validation
+{
+    public class LongRunningOperationsOptionsValidator : TypedRule<Dictionary<string, Operation>>
```
veronicagg: is there a reason to implement the rule in the .net validator instead of the typescript validator?

veronicagg: if in .net, should the rule be in OpenAPI.Validator.Validation.Extensions namespace, like https://github.com/Azure/azure-openapi-validator/blob/401cf16dd477093a4edef2ba57be71a2f279ea34/src/dotnet/OpenAPI.Validator/Validation/Extensions/LongRunningResponseStatusCode.cs#L13

src/dotnet/OpenAPI.Validator/Validation/LongRunningOperationsOptionsValidator.cs
```diff
@@ -0,0 +1,89 @@
...
+        /// <summary>
+        /// The rule could be violated by a model referenced by many jsons belonging to the same
+        /// composed state, to reduce duplicate messages, run validation rule in composed state
+        /// </summary>
+        public override ServiceDefinitionDocumentState ValidationRuleMergeState => ServiceDefinitionDocumentState.Individual;
+
+        /// <summary>
+        /// Validates if a Post LRO Operation with return value has long running operation options extensions enabled.
+        /// </summary>
+        /// <param name="operation"></param>
```
veronicagg: update parameter comment

src/dotnet/OpenAPI.Validator/Validation/LongRunningOperationsOptionsValidator.cs
```diff
@@ -0,0 +1,89 @@
...
+        /// <param name="context"></param>
+        /// <returns>true if a Post LRO Operation with return value has long running operation options extensions enabled. false otherwise.</returns>
+        public override IEnumerable<ValidationMessage> GetValidationMessages(Dictionary<string, Operation> operationPaths, RuleContext context)
+        {
+            Operation operation = null;
+            // Check if there is a post operation. We are not considering any other operation for this rule.
+            if (operationPaths.TryGetValue("post", out operation))
```
veronicagg: what comes in operationsPaths? could there be multiple POST operations?

sarangan12: No. It cannot have more than one post. Autorest will crash

src/dotnet/OpenAPI.Validator/Properties/Resources.resx
sarangan12: Changed to return schema. Will add the documentation. Such a message is not necessary because operation will be post which you are already indication and return schema is not important.

src/dotnet/OpenAPI.Validator/Validation/LongRunningOperationsOptionsValidator.cs
sarangan12: Done

src/dotnet/OpenAPI.Validator/Validation/LongRunningOperationsOptionsValidator.cs
sarangan12: Done

sarangan12:
Doc added: https://github.com/Azure/azure-rest-api-specs/pull/4039

veronicagg:
@sarangan12 what's the reason to have implemented the rule in the .Net validator instead of the typescript one?
@lmazuel would you like to review the change too?

sarangan12:
@veronicagg Just personal choice. No specific reason.

PR #181, opened 2018-10-11
Add response for 201

zjpjack:
Add 201 response

### DIFF
--- a/src/dotnet/OpenAPI.Validator.Tests/Resource/OpenAPI/Validation/tracked-resource-patch-operation.json
+++ b/src/dotnet/OpenAPI.Validator.Tests/Resource/OpenAPI/Validation/tracked-resource-patch-operation.json
@@ -10,8 +10,12 @@
     "https"
   ],
   "basePath": "/",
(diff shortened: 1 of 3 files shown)


src/dotnet/OpenAPI.Validator.Tests/Resource/OpenAPI/Validation/tracked-resource-patch-operation.json
```diff
@@ -20,8 +24,7 @@
         ],
         "operationId": "Redis_Get",
         "description": "gets a Redis cache.",
-        "parameters": [
-        ],
+        "parameters": [],
         "responses": {
           "200": {
```
dsgouda: You should probably create a positive new test with a new test json which has `201` response as expected. This file should remain unchanged

zjpjack:
@dsgouda I checked CI result is succeed, why it is still pending here

dsgouda:
Hmm... for some reason travis did not report it on the PR

PR #189, opened 2019-12-24
Change azure pipeline

qiaozha:


src/typescript/azure-openapi-validator/rule.ts
```diff
@@ -21,6 +21,7 @@ interface ValidationMessage {
 }
 
 export interface Rule {
+
```
ruowan: Why the empty line here?

azure-pipelines/ci.yml
```diff
@@ -0,0 +1,57 @@
+trigger:
```
ruowan: you can directly use azure-pipelines.yml in root folder level instead of putting it into a folder.

qiaozha: to test pipeline

qiaozha: I want to create a release pipeline later to separate it

src/typescript/azure-openapi-validator/rule.ts
qiaozha: to test pipeline

azure-pipelines/ci.yml
ruowan: For release pipeline. It's private you can't put it into here.

qiaozha: I mean in a separate file but in azure-pipelines folder. That's why I create a folder. And the release pipeline will sure be put in internal pipelines.

src/typescript/azure-openapi-validator/rule.ts
ruowan: For testing pipeline. You can comments /azp run.  So you can revert the changes, it's not necessary.

azure-pipelines/ci.yml
ruowan: Release pipeline YAML  file can't be put into GitHub repo.   It is stored in DevOps. So this repo only contains one pipeline. Create a folder will make it no clear.

src/typescript/azure-openapi-validator/rule.ts
qiaozha: changed

azure-pipelines/ci.yml
qiaozha: changed

PR #192, opened 2020-02-14
updated R3030: the last resource provider matches namespace

ruowan:


src/dotnet/OpenAPI.Validator.Tests/OpenAPIModelerValidationTests.cs
```diff
@@ -699,6 +699,7 @@ public void CleanFileValidation()
                 MergeState = ServiceDefinitionDocumentState.Individual
             };
             var messages = this.ValidateOpenAPISpec(Path.Combine(AutoRest.Core.Utilities.Extensions.CodeBaseDirectory, "Resource", "OpenAPI", "Validation", "positive", "clean-complex-spec.json"), subtest1md);
+
```
lirenhe: Do you have any test case to cover this?

lirenhe: Do you have any test case to cover this?

src/dotnet/OpenAPI.Validator/Model/Utilities/ValidationUtilities.cs
```diff
@@ -20,6 +20,8 @@ public static class ValidationUtilities
         private static readonly Regex ResourceProviderPathPattern = new Regex(@"/providers/(?<resPath>[^{/]+)/", RegexOptions.IgnoreCase);
         private static readonly Regex PropNameRegEx = new Regex(@"^[a-z0-9\$-]+([A-Z]{1,3}[a-z0-9\$-]+)+$|^[a-z0-9\$-]+$|^[a-z0-9\$-]+([A-Z]{1,3}[a-z0-9\$-]+)*[A-Z]{1,3}$");
 
+        private static readonly Regex ResourceProviderNamespaceRegex = new Regex(@"(resource-manager|data-plane)/(?<namespace>[\w\.]+)/");
```
lirenhe: For this case, we are not updating an existing rule, actually, we are retiring R2059 and use a new rule instead.
What if we have a new rule id for this?

src/dotnet/OpenAPI.Validator.Tests/OpenAPIModelerValidationTests.cs
ruowan: Added unittest.

ruowan: Add unittest

src/dotnet/OpenAPI.Validator/Model/Utilities/ValidationUtilities.cs
josefree: Agree with Renhe, it seems to be a new rule, instead of upgrade the original rule R2025.  How about we set up a new rule?

src/dotnet/OpenAPI.Validator/Properties/Resources.resx
```diff
@@ -208,7 +208,7 @@
     <value>Parameter "{0}" is referenced but not defined in the global parameters section of Service Definition</value>
   </data>
   <data name="UniqueResourcePaths" xml:space="preserve">
```
lirenhe: Could you also update the 'data name'

ruowan: updated.

PR #193, opened 2020-02-18
Ruowan/add emum rule

ruowan:


changelog.md
```diff
@@ -1,8 +1,15 @@
 # Changelog
 
+## What's New (2/18/2020)
```
lirenhe: You may work with Jianye for the complete change log. I think we will only release one that combines the change from you and Jianye.

src/typescript/azure-openapi-validator/rules/EnumMustHaveType.ts
```diff
@@ -0,0 +1,22 @@
...
+  id: "R3015",
+  name: EnumMustHaveType,
+  severity: "warning",
+  category: "ARMViolation",
+  mergeState: MergeStates.individual,
+  openapiType: OpenApiTypes.arm | OpenApiTypes.dataplane,
+  appliesTo_JsonQuery: "$..*[?(@.enum)]",
+  run: function*(doc, node, path) {
+    const msg: string = `Enum must define its type and "object" type is not allowed due to Autorest refuse to parse it.`;
+    if (node.type === undefined || node.type.toLowerCase() === "object") {
```
lirenhe: could you use !node.type instead?

src/typescript/azure-openapi-validator/rules/EnumUniqueValue.ts
```diff
@@ -0,0 +1,34 @@
...
+rules.push({
+  id: "R3024",
+  name: EnumUniqueValue,
+  severity: "warning",
+  category: "ARMViolation",
+  mergeState: MergeStates.individual,
+  openapiType: OpenApiTypes.arm | OpenApiTypes.dataplane,
+  appliesTo_JsonQuery: "$..*[?(@.enum)]",
+  run: function*(doc, node, path) {
+    const msg: string = `Enum must not contain case-insensitive duplicated value and make sure every value in enum unique.`;
```
lirenhe: could you update the message to contain enum value?

src/typescript/package.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "@microsoft.azure/openapi-validator",
-    "version": "1.0.5",
+    "version": "1.2.0",
```
lirenhe: Please work with Jianye to change the version together and why use 1.2.0 directly?

changelog.md
ruowan: Ok. I will talk about Jianye offline.

src/typescript/package.json
ruowan: Ok. We will create a separate PR to bump version and update changelog.

src/typescript/azure-openapi-validator/rules/EnumMustHaveType.ts
ruowan: I perfer ”node.type===undefined“ is more clear and means author forget to define its type. `!node.type` include `false, empty, null...`.

src/typescript/azure-openapi-validator/rules/EnumUniqueValue.ts
ruowan: Updated.

src/typescript/azure-openapi-validator/rules/EnumMustHaveType.ts
lirenhe: what is the type of node.type, is it string?
what if node.type is null?

ruowan: type is string.

PR #195, opened 2020-02-21
Update R3006,R2016

jianyexi:


src/dotnet/OpenAPI.Validator.Tests/OpenAPIModelerValidationTests.cs
```diff
@@ -907,6 +915,13 @@ public void ValidXMSPathWithOAata()
             }
             Assert.Empty(message);
         }
+
+        [Fact]
+        public void ValidRequiredDesciminatorPropertiesInPatchRequestValidation()
```
lirenhe: do you have the existing sample to verify R2016? I don't think we need to create a new test case for this.

src/dotnet/OpenAPI.Validator/Model/Utilities/ValidationUtilities.cs
```diff
@@ -169,6 +169,25 @@ public static IEnumerable<string> EnumerateRequiredProperties(string modelName,
...
+        /// <param name="modelName">model for which to check the discrimintor properties</param>
+        /// <param name="definitions">dictionary of model definitions</param>
+        /// <param name="propertyList">List of required properties found in model hierarchy</param>
+        public static IEnumerable<string> EnumerateDiscrimintorProperties(string modelName, Dictionary<string, Schema> definitions)
+        {
+            var modelsToCheck = EnumerateModelHierarchy(modelName, definitions);
+            var propertiesList = new List<string>();
+            foreach (var modelRef in modelsToCheck)
+            {
+                if (!definitions.ContainsKey(modelRef) || definitions[modelRef].Discriminator?.Any() != true) continue;
```
lirenhe: could you refactor this code?  no need to combine them in one line.

src/dotnet/OpenAPI.Validator/Model/Utilities/ValidationUtilities.cs
```diff
@@ -169,6 +169,25 @@ public static IEnumerable<string> EnumerateRequiredProperties(string modelName,
             return propertiesList;
         }
 
+        /// <summary>
+        /// Returns the cumulative list of all discrimintor properties found in the entire model hierarchy
+        /// </summary>discrimintor
+        /// <param name="modelName">model for which to check the discrimintor properties</param>
+        /// <param name="definitions">dictionary of model definitions</param>
+        /// <param name="propertyList">List of required properties found in model hierarchy</param>
```
lirenhe: is this parameter used?

jianyexi: thank you, I will remove it

src/dotnet/OpenAPI.Validator.Tests/OpenAPIModelerValidationTests.cs
jianyexi: this case is a litter special, so I add a new to test it
Btw I think we should modify one test case only when it has some bug

src/dotnet/OpenAPI.Validator/Model/Utilities/ValidationUtilities.cs
jianyexi: this '!= true' can not be replaced by '!', or will compile failed

src/dotnet/OpenAPI.Validator.Tests/OpenAPIModelerValidationTests.cs
lirenhe: Not necessary. But anyway, you could make the decision for this.

src/dotnet/OpenAPI.Validator/Model/Utilities/ValidationUtilities.cs
```diff
@@ -169,6 +169,26 @@ public static IEnumerable<string> EnumerateRequiredProperties(string modelName,
...
+        /// </summary>discrimintor
+        /// <param name="modelName">model for which to check the discrimintor properties</param>
+        /// <param name="definitions">dictionary of model definitions</param>
+        public static IEnumerable<string> EnumerateDiscrimintorProperties(string modelName, Dictionary<string, Schema> definitions)
+        {
+            var modelsToCheck = EnumerateModelHierarchy(modelName, definitions);
+            var propertiesList = new List<string>();
+            foreach (var modelRef in modelsToCheck)
+            {
+                if (!definitions.ContainsKey(modelRef) || definitions[modelRef].Discriminator?.Any() != true) {
```
lirenhe: Could you use '!definitions[modelRef].Discriminator?.Any()' directly?

ruowan: Why use Any()? use String.IsNullOrEmpty(String)  is more clear for string type.

jianyexi: Good suggestion!

PR #197, opened 2020-02-25
add pre release pipeline

jianyexi:


azure-pipelines/pre-release-pipelines.yml
```diff
@@ -0,0 +1,78 @@
...
+        inputs:
+          pathtoPublish: $(Build.SourcesDirectory)/dist
+
+  - job: Windows
+    displayName: credentialScan
+    pool:
+      vmImage: "windows-2019"
+
+    steps:
+      - task: ea576cd4-c61f-48f8-97e7-a3cb07b90a6f@2 # CredScan@2
```
lirenhe: Could you use a user friendly name for this task?

azure-pipelines/pre-release-pipelines.yml
```diff
@@ -0,0 +1,78 @@
...
+    displayName: credentialScan
+    pool:
+      vmImage: "windows-2019"
+
+    steps:
+      - task: ea576cd4-c61f-48f8-97e7-a3cb07b90a6f@2 # CredScan@2  
+        inputs:
+          toolMajorVersion: 'V2'
+
+      - task: f5679091-e6da-4974-a8dc-0eec03a8ea63@1 # PostAnalysis@1
```
lirenhe: Same above

azure-pipelines/pre-release-pipelines.yml
```diff
@@ -0,0 +1,78 @@
...
+      - script: 'gulp build'
+        displayName: 'gulp build'
+
+      - script: 'gulp test'
+        displayName: 'gulp test'
+        
+      - script: 'gulp dotnet/pack'
+        displayName: 'gulp dotnet/pack'
+        
+      - task: EsrpCodeSigning@1
```
lirenhe: How is this pipeline triggered? As we discussed last time, we don't auto sign the code.

azure-pipelines/pre-release-pipelines.yml
jianyexi: there is a issue in DevOps  cause the task name is always ambiguous , so here,  provided a guid instead of task name to avoid it

azure-pipelines/pre-release-pipelines.yml
jianyexi: Since this pipeline is internal ，we can disabled CI and  trigger it manually

lirenhe: Who has the permission to trigger this? Any approve needed?

jianyexi: only project azure-sdk  members can trigger it，currently there is no approve process  in build pipeline  it has in release
 Pipeline ,but it’s safe , because
 it’s internal，and we can know who triggered the pipeline from the history log

PR #209, opened 2020-05-20
Update r4005 

jianyexi:


jianyexi:
It's allowed to have two enums  defined in one swagger which have the same name and same entries  ,

PR #212, opened 2020-06-23
update gulp version to fit node v12

jianyexi:


src/typescript/package.json
```diff
@@ -25,25 +25,24 @@
...
+        "gulp-dotnet-cli": "1.1.0",
         "mocha-typescript": "1.0.22",
         "typescript": "3.5.3"
     },
     "dependencies": {
         "fs": "^0.0.1-security",
-        "js-yaml": "^3.8.4",
-        "jsonpath": "^0.2.11",
+        "js-yaml": "^3.14.0",
+        "jsonpath": "^1.0.2",
```
ruowan, line 42: Is there any breaking change for updating it?

jianyexi: Have no effect with the Rules, just update them to more stable version

gulpfile.js
```diff
@@ -53,63 +54,89 @@ gulp.task('clean/dotnet', function () {
...
-gulp.task('pack/typescript', [], function () {
+gulp.task(
+  "build",
+  gulp.parallel(["build/dotnet", "build/typescript"], function (done) {
+    console.log("Building code...")
+    done()
+  })
+)
+
+gulp.task(
```
ruowan, line 84: Why you add gulp.parallel and gulp.series? what do they do?

jianyexi: The two fucntions is required in  new version of gulp, they determin the task will  run series or parallel

PR #213, opened 2020-06-23
add new openAPI Type:Rpass

jianyexi:


src/typescript/index.ts
```diff
@@ -14,6 +14,8 @@ async function main() {
     const files = await initiator.ListInputs()
     const mergeState: string = await initiator.GetValue("merge-state")
     const openapiType: string = await initiator.GetValue("openapi-type")
+    const subType: string = await initiator.GetValue("openapi-subtype")
```
ruowan, line 17: subType means?

jianyexi: Currently used by rpass and passed through by autorest .  If existed,  we only need to run the rpass rules

PR #216, opened 2020-07-30
Hot fix

jianyexi:
1 R4007 throw exception when examples contains $ref


src/typescript/azure-openapi-validator/rules/utilities/rules-helper.ts
```diff
@@ -53,14 +53,21 @@ export async function getResolvedJson(doc: any): Promise<object | undefined> {
   try {
     const parser = new $RefParser()
     const docCopy = JSON.parse(JSON.stringify(doc))
+    /*
+     * remove x-ms-examples
+     */
+    apply(docCopy, `$..['x-ms-examples']`, e => null)
```
ruowan, line 59: what does this function do?

src/typescript/azure-openapi-validator/tests/resources/utilities/exception.json
```diff
@@ -0,0 +1,1177 @@
+{
```
ruowan, line 1: prefer to use small example and change file name `exception.json` more specific.

src/typescript/azure-openapi-validator/rules/utilities/rules-helper.ts
jianyexi: set x-ms-example to null , as we don't need to check it

PR #233, opened 2020-11-30
update rules & add changlog

jianyexi:


ruowan (commented):
Do we need to bump verison in the PR?

jianyexi:
> Do we need to bump verison in the PR?

Good point, already bumped

ruowan (approved):

PR #251, opened 2021-11-10
add new rule

tianxchen-ms:
Enum values should respect the type specifier.


src/typescript/azure-openapi-validator/rules/EnumMustRespectType.ts
```diff
@@ -0,0 +1,36 @@
...
+      if (node.enum && isValidEnum(node)) {
+        if (
+          node.enum.some(value => {
+            if (node.type === "integer" && !Number.isInteger(value)) {
+              return true
+            }
+            return typeof value !== node.type
+          })
+        ) {
+          yield { message: `${msg}`, location: path }
```
jianyexi, line 31: the location should be more accurate here,  may use  `result.path`  to instead

jianyexi (approved):

PR #261, opened 2022-01-25
Fix R4011 (DeleteOperationResponses) to accept either a 200 or 204 response

smichtch:
Existing rule implementation requires `DELETE` endpoints to include both a 200 and a 204 response.


smichtch:
Updated the regression snapshot, spot-checked a few removed instances of `DeleteOperationResponses` and it looks accurate.
The only remaining `DeleteOperationResponses` instance looks legitimate:
https://github.com/Azure/azure-rest-api-specs/blob/fce3400431eff281bddd04bed9727e63765b8da0/specification/powerbiembedded/resource-manager/Microsoft.PowerBI/stable/2016-01-29/powerbiembedded.json#L220-L229

jianyexi:
Great,  could you update it to include '202' , as it's ok to only return 202 for long running

PR #292, opened 2022-05-07
add new rules

jianyexi:
add rules :

1. https://github.com/Azure/azure-openapi-validator/issues/303
   If a parameter matches belows, therefore it must be called 'endpoint' and be typed 'type:string, format:uri'.
    - Client level (x-ms-parameter-location: client)
    - A path component (in: path)
    - Part of a 'x-ms-parametrized-host' with 'useSchemePrefix: false'
    - Tagged 'x-ms-skip-encoding: true'
3. https://github.com/Azure/azure-openapi-validator/issues/305

4. https://github.com/Azure/azure-openapi-validator/issues/301
5. https://github.com/Azure/azure-openapi-validator/issues/304


docs/host-parameters-validation.md
```diff
@@ -0,0 +1,19 @@
+# HostParametersValidation
+
+## Description
+
+This rule is to valiate the parameters in the 'x-ms-parameterized-host' to ensure they are following below rules:
```
josefree, line 5: This is to validate if parameters in the 'x-ms-parameterized-host' follow the following rules:

docs/host-parameters-validation.md
```diff
@@ -0,0 +1,19 @@
...
+
+1. If a parameter matches belows, therefore it must be called 'endpoint' and be typed 'type:string, format:uri'.
```
josefree, line 7: If a parameter matches one of the following cases, it must be defined as 'endpoint' and must be claimed with 'type:string, format:uri' explicitly.

docs/lro-post-must-not-use-original-url-as-final-state.md
```diff
@@ -0,0 +1,13 @@
+# LroPostMustNotUseOriginalUriAsFinalState
+
+## Description
+
+The long running post operation must not use final-stat-via:original-uri.
+
+## How to fix
+
+Change to use other options define in [x-ms-long-running-operation-options](https://github.com/Azure/autorest/tree/main/docs/extensions#x-ms-long-running-operation-options)
```
josefree, line 9: Consider using states defined in [x-ms-long-running-operation-options](https://github.com/Azure/autorest/tree/main/docs/extensions#x-ms-long-running-operation-options)

docs/lro-with-original-url-as-final-state.md
```diff
@@ -0,0 +1,13 @@
...
+
+The long running operation with final-state-via:original-uri should have a sibling 'get' operation.
+
+## How to fix
+
+Add the missing get operation.
+
+## What's the impact if breaking the rule
+
+If no get operation, the generated SDK could not know the response schema of the final state.
```
josefree, line 13: If get operation is missing, SDK generation may fail for the response schema of the final state is not available.

docs/patch-body-parameters-schema.md
```diff
@@ -0,0 +1,52 @@
+# PatchBodyParametersSchema
+
+## Description
+
+ A request parameter of the Patch Operation must not have a required/default/'x-ms-mutability: ["create"]' value.
+
+## Why the rule is important
+
+ A PATCH operation is used to update properties of a resource. So, If the resource has 'X' number of properties and if you wish to change one of them, then a PATCH request could be sent with a value for that specified property. In other words, all the properties in the PATCH request are updated. Now, if any of the values are marked as required/default, it would force the system to update it always which is not the intention of the PATCH operation.
```
josefree, line 9: A PATCH operation is used to update properties of a resource.  Therefore, if the resource has 'X' number of properties and the purpose is to update one of them, then a PATCH request could be sent with a value for that specified property. Considering all properties in the PATCH request are to be updated, if any of values in PATCH request are marked as required/default, the updates are forced to happen although it may not be the intention of the PATCH operation.

docs/rules.md
```diff
@@ -10,6 +10,12 @@ This rule appears when you define a model type inline, rather than in the `defin
 
 Please refer to [anonymous-parameter-types.md](./anonymous-parameter-types.md) for details.
 
+### ArrayMustHaveType
+
+Array type must have a type except for any type.
```
josefree, line 15: An array must have "type:'array'" in schema explicitly.

josefree (changes requested):

docs/host-parameters-validation.md
jianyexi: "it must be defined as 'endpoint'"  seems a little confusion, here it mean the parameter name must be 'endpoint'

docs/lro-with-original-url-as-final-state.md
jianyexi: seem a little confusion for me, actually I can tell more here, if no get operation, SDK have no way to determine the response schema of the final state  , as it 's same with get operation schema

josefree (approved):

xiaoxuqi-ms (approved):

PR #338, opened 2022-06-19
Add docs for Azure data plane rules

mikekistler:
This is a draft PR containing documentation for one data plane Spectral rule. I'd like confirmation that this is the right location and format of the docs before proceeding on the remaining rules.


docs/additional-properties-and-properties.md
```diff
@@ -0,0 +1,29 @@
+# AdditionalPropertiesAndProperties
+
+## Category
+
+Data Plane Info
```
jianyexi, line 5: Seem we only have ''Error" and "Warning", what do you expect user to do with an 'Info' message ?

jianyexi:
> This is a draft PR containing documentation for one data plane Spectral rule. I'd like confirmation that this is the right location and format of the docs before proceeding on the remaining rules.

yes, you can run 'rush regen-ruleindex' to update the  'rules.md' , currently it's only used to index all the separate rule file

docs/additional-properties-and-properties.md
mikekistler: In the Spectral ruleset, we use 'info' to flag a case that might not be wrong, but should be given some closer scrutiny.  Many of our Spectral rules are 'info' severity.  But I can use "Warning" as long as that does not block a PR merge.

mikekistler:
For this PR, which focuses only on documentation, I can make the names match the names in the rule file, and then we can update the rule names in a future PR.  Sound reasonable?

jianyexi (approved):

Issue #299, opened 2022-05-11
[new rule]: Information in the URI should not be repeated in the request body


jianyexi:
### Lint rule description

Information in the URI should not be repeated in the request body (i.e. subscription ID) 
this rule is to check the properties under 'properties', as we already have rule to check the top level property 

### Related swagger example

N/a

### Category

ARM

### Severity level

Error

### Applies to

Management plane API spec

### How to fix the violation

remove the repeated properties like 'subscriptionId' or 'resourceGroup', 'resourceName' from the properties bag

### What't the impact if breaking the rule

Repetitive info and too easy to get out of sync during moves

Issue #298, opened 2022-05-11
[new rule]: URI for put must be under a subscription and resource group 


jianyexi:
### Lint rule description

URI must be under a subscription and resource group 
for a PUT operation, If a uri segment has subscription, it needs to have a resourcegroup segment as well 

### Related swagger example

   "/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.AzureData/sqlServerRegistrations/{sqlServerRegistrationName}"

### Category

ARM

### Severity level

Error

### Applies to

Management plane API spec

### How to fix the violation

add the resource group or suscription segment and parameters

### What't the impact if breaking the rule

does not conform to ARM spec

Issue #296, opened 2022-05-11
[new rule]: Uri for resource action method (POST) MUST specify a resource and end with an action name


jianyexi:
### Lint rule description

Uri for resource action method (POST) MUST specify a resource and end with an action name

### Related swagger example

"/{scope}/providers/Microsoft.Authorization/roleAssignmentApprovals/{approvalId}/stages"

### Category

ARM

### Severity level

Error

### Applies to

Management plane API spec

### How to fix the violation

Uri path specification follows <scope>/providers/\<namespace\>/\<resource\>/\<actionname\> format, where \<scope\>, \<namespace\>, and \<resource\>,  are as specified in RPC-V2-URI-6. The \<actionname\> segment MUST be a literal. Rules for action names: 

MUST be a word or words that describe the action to be taken on a single resource. 

### What't the impact if breaking the rule

does not conform to ARM RPC

Issue #295, opened 2022-05-11
[new rule]: Uri for CRUD methods on a nested resource type MUST follow valid resource naming 


jianyexi:
### Lint rule description

Uri for CRUD methods on a nested resource type MUST follow valid resource naming 

### Related swagger example

/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.DocumentDB/resourceType/{resourceName}/subResourceType/{subResoureName}

### Category

ARM

### Severity level

Error

### Applies to

Management plane API spec

### How to fix the violation

Uri path specification starts with <scope>/providers/<namespace>/<resource> format,

Rules for the <resource> segment(s): 

MUST follow format of <typename>/{resourceName}/<descendents> where <descendents> follows the format of <subtypename>/{subResourceName} 

<typename> and <subtypename> must be literal resource type names following RPC-V2-URI-4. 

{resourceName} MUST be a parameter named "resourceName". Descendent resource name parameters MUST be unique and correspond to their subtype. 

### What't the impact if breaking the rule

Does not conform to ARM RPC

PR #348, opened 2022-06-24
Arm rules regarding to Uri

jianyexi:
add below rules

- [x] https://github.com/Azure/azure-openapi-validator/issues/283

- [x] https://github.com/Azure/azure-openapi-validator/issues/284

- [x] https://github.com/Azure/azure-openapi-validator/issues/285

- [x] https://github.com/Azure/azure-openapi-validator/issues/295

- [x] https://github.com/Azure/azure-openapi-validator/issues/296

- [x] https://github.com/Azure/azure-openapi-validator/issues/298

- [x] https://github.com/Azure/azure-openapi-validator/issues/299


xiaoxuqi-ms (approved):

xiaoxuqi-ms (approved):

packages/rulesets/src/spectral/functions/arm-path-validation.ts
```diff
@@ -0,0 +1,167 @@
...
+
+function verifyResourceGroupScope(path: string) {
+  // valid patterns:
+  // 1 /<scope>/providers/Microsoft.Compute/virtualMachine/{vmName}
+  // 2 /subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.Compute/virtualMachine/{vmName}
+  // 3 /subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.Compute/virtualMachine/{vmName}/providers/Microsoft.Billing/extensions/{extensionName}
+  const patterns = [
+    /^\/subscriptions\/{subscriptionId}\/resourceGroups\/{resourceGroupName}\/providers\/.+/gi,
+    /^\/?{\w+}\/resourceGroups\/{resourceGroupName}\/providers\/.+/gi,
+    /^\/?{\w+}\/providers\/.+/gi,
```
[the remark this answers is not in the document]
jianyexi: here , I want to match '/{scope}/provider...'  which is valid pattern 1 in the annotation

packages/rulesets/src/spectral/functions/arm-path-validation.ts
```diff
@@ -0,0 +1,167 @@
...
+  return matchAnyPatterns(patterns, path)
+}
+
+function verifyNestResourceType(path: string) {
+  // invalid patterns:
+  // 1 <scope>/providers/Microsoft.Compute/virtualMachine/{vmName}/nestedResourceType/actions
+  // 2 <scope>/providers/Microsoft.Compute/virtualMachine/{vmName}/{nestedResourceType}
+
+  const patterns = [
+    /^.*\/providers\/microsoft\.\w+\/\w+\/{\w+}(?:\/\w+\/(?!default)\w+){1,2}/gi,
```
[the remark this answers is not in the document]
jianyexi: it's used to match multiple times of the previous pattern

packages/rulesets/src/spectral/az-arm.ts
```diff
@@ -96,6 +98,102 @@ const ruleset: any = {
...
+        },
+      },
+    },
+    URIForResourceAction: {
+      description: "Uri for 'post' method on a resource type MUST follow valid resource naming.",
+      message: "{{description}}",
+      severity: "warn",
+      resolved: false,
+      formats: [oas2],
+      given: ["$[paths,'x-ms-paths'].*.post^~"],
```
mikekistler, line 177:
```suggestion
      given: "$[paths,'x-ms-paths'].*.post^~",
```

packages/rulesets/src/spectral/az-arm.ts
```diff
@@ -96,6 +98,102 @@ const ruleset: any = {
...
+      },
+    },
+    RepeatedUriInfo: {
+      description:
+        "Information in the URI should not be repeated in the request body (i.e. subscription ID, resource group name, resource name).",
+      message: "The '{{error}}' already appears in the URI, please don't repeat it in the request body.",
+      severity: "warn",
+      resolved: true,
+      formats: [oas2],
+      given: ["$[paths,'x-ms-paths'].*.put^"],
```
mikekistler, line 192:
```suggestion
      given: "$[paths,'x-ms-paths'].*.put^",
```

packages/rulesets/src/spectral/az-arm.ts
```diff
@@ -96,6 +98,102 @@ const ruleset: any = {
         function: falsy,
       },
     },
+    URIContainsSubscriptionId: {
+      description: "Uri for resource group scoped CRUD methods MUST contain a subscriptionId parameter.",
+      message: "{{error}}",
+      severity: "error",
+      resolved: false,
```
mikekistler, line 105: Why do you need `resolved: false` here? You are using `resolved: true` for other rules that use the same `verifyArmPath` function, so I think this is probably not needed (or wanted).

packages/rulesets/src/spectral/az-arm.ts
```diff
@@ -96,6 +98,102 @@ const ruleset: any = {
...
+        functionOptions: {
+          segmentToCheck: "subscriptionIdParam",
+        },
+      },
+    },
+    URIContainsResourceType: {
+      description: "Uri for resource CRUD methods MUST contain a resource type.",
+      message: "{{error}}",
+      severity: "error",
+      resolved: true,
```
mikekistler, line 119: `resolved: true` is the default so you can just omit it.

packages/rulesets/src/spectral/az-arm.ts
```diff
@@ -96,6 +98,102 @@ const ruleset: any = {
         function: falsy,
       },
     },
+    URIContainsSubscriptionId: {
```
mikekistler, line 101: I think all these rules should be renamed to "PathContains..." since they check the "path" element.  Yes, the path will be used as a URL in a request, but we should match the terminology with features of the API definition that we are actually checking.

packages/rulesets/src/spectral/functions/arm-path-validation.ts
```diff
@@ -0,0 +1,167 @@
...
+  const path = paths.path || []
+
+  const errors: any[] = []
+
+  const optionsHandlers = {
+    resourceType: (pathKey: string) => {
+      if (!verifyResourceType(pathKey)) {
+        errors.push({
+          message: `The URI for the CURD methods do not contain a resource type.`,
+          path: [...path],
```
mikekistler, line 115: I think you can just use
```suggestion
          path,
```

packages/rulesets/src/spectral/functions/arm-path-validation.ts
```diff
@@ -0,0 +1,167 @@
...
+  ) {
+    return []
+  }
+  const path = paths.path || []
+
+  const errors: any[] = []
+
+  const optionsHandlers = {
+    resourceType: (pathKey: string) => {
+      if (!verifyResourceType(pathKey)) {
```
mikekistler, line 112: This does not make sense to me. Here you are calling `verifyResourceType` with a "pathKey", but the `verifyResourceType` function accepts a "path". This might be just a naming thing but it is confusing. What is being passed here -- the full path or just a path segment?

The same on all the other functions here.

packages/rulesets/src/spectral/functions/arm-path-validation.ts
```diff
@@ -0,0 +1,167 @@
...
+  if (pathKey === null || typeof pathKey !== "string") {
+    return []
+  }
+
+  const validOptions = ["scope", "resourceGroupParam", "subscriptionIdParam", "resourceGroupScope", "resourceType", "nestedResourceType"]
+  if (
+    !_opts ||
+    !_opts.segmentToCheck ||
+    (typeof _opts.segmentToCheck === "string" && !validOptions.includes(_opts.segmentToCheck)) ||
+    (Array.isArray(_opts.segmentToCheck) && !_opts.segmentToCheck.every((s: string) => validOptions.includes(s)))
```
mikekistler, line 102: I think Spectral lets you define very precisely what the shape of `_opts` must be and that would let you avoid a bunch of the logic here.

packages/rulesets/src/spectral/functions/arm-path-validation.ts
```diff
@@ -0,0 +1,167 @@
...
+      }
+    },
+    scope: (pathKey: string) => {
+      if (!verifyScope(pathKey)) {
+        errors.push({
+          message: "",
+          path: [...path],
+        })
+      }
+    },
```
mikekistler, line 150: It looks like this function is unused.

packages/rulesets/src/spectral/functions/arm-path-validation.ts
```diff
@@ -0,0 +1,167 @@
...
+  // 1 /<scope>/providers/Microsoft.Compute/virtualMachine/{vmName}
+  // 2 /subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.Compute/virtualMachine/{vmName}
+  // 3 /subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.Compute/virtualMachine/{vmName}/providers/Microsoft.Billing/extensions/{extensionName}
+  const patterns = [
+    /^\/subscriptions\/{subscriptionId}\/resourceGroups\/{resourceGroupName}\/providers\/.+/gi,
+    /^\/?{\w+}\/resourceGroups\/{resourceGroupName}\/providers\/.+/gi,
+    /^\/?{\w+}\/providers\/.+/gi,
+  ]
+  return matchAnyPatterns(patterns, path)
+}
```
mikekistler, line 44: It looks like this function is unused.  Do we need it?

mikekistler (changes requested):
I left a bunch of comments. I hope these are helpful, but some may be just my lack of understanding of how the Spectral rules work inside aov.

packages/rulesets/src/spectral/az-arm.ts
jianyexi: I'd like to call it 'URI'  because actually what I want to ensure is the URI not 'path', for example the path may not contains 'subscriptionId' instead it contains '{scope}', that's ok

packages/rulesets/src/spectral/az-arm.ts
jianyexi: if for verifying swagger path only, we can just lint on the un-resolved swagger, that would be more efficient

mikekistler: I don't believe you gain any efficiency in linting the unresolved API definition.  I believe Spectral _always_ computes the resolved doc _once_, before running any rules, and `resolved` simply determines which input is passed to the function.

Do you have performance measurements that suggest otherwise?

packages/rulesets/src/spectral/az-arm.ts
mikekistler: I still think "Path" is the correct term here. Looking at the existing rules, there are none whose name starts with "URI", but several that start with "Path", e.g. "PathResourceProviderMatchNamespace", "PathResourceProviderNamePascalCase", "PathResourceTypeNameCamelCase".  So on precedent alone I think we should be using "Path".

And reiterating my earlier point -- this tool checks an OpenAPI v2 document for compliance to our guidelines. We should use the terminology of OpenAPI v2 to describe problems.  There is not one mention of the term "URI" in the OpenAPI v2 spec. The element in the doc that you flag as in error is a "path". Why not call it a "Path"?

packages/rulesets/generated/spectral/az-arm.js
```diff
@@ -259,6 +407,101 @@ const ruleset = {
...
+                },
+            },
+        },
+        URIForResourceAction: {
+            description: "Uri for 'post' method on a resource type MUST follow valid resource naming.",
+            message: "{{description}}",
+            severity: "warn",
+            resolved: false,
+            formats: [oas2],
+            given: ["$[paths,'x-ms-paths'].*.post^~"],
```
mikekistler, line 486: one more
```suggestion
            given: "$[paths,'x-ms-paths'].*.post^~",
```

packages/rulesets/generated/spectral/az-arm.js
```diff
@@ -259,6 +407,101 @@ const ruleset = {
...
+                },
+            },
+        },
+        RepeatedUriInfo: {
+            description: "Information in the URI should not be repeated in the request body (i.e. subscription ID, resource group name, resource name).",
+            message: "The '{{error}}' already appears in the URI, please don't repeat it in the request body.",
+            severity: "warn",
+            resolved: true,
+            formats: [oas2],
+            given: ["$[paths,'x-ms-paths'].*.put^"],
```
mikekistler, line 500: one more 
```suggestion
            given: "$[paths,'x-ms-paths'].*.put^",
```

packages/rulesets/src/spectral/az-arm.ts
mikekistler: @rkmanda Do you have a preference for "URI" vs "Path"? Since these are rules specifically for ARM we should probably let you decide which you think the ARM reviewers will prefer.

rkmanda: I have no preference. Good with path.

jianyexi:
already replaced 'uri' with 'path'

PR #359, opened 2022-07-18
Update CONTRIBUTING.md

mikekistler:
Updated prerequisites in CONTRIBUTING.md.
- Updated required version of Node as Node 12 is EOL.
- Removed npm from prereqs as it now comes with node.
- Removed typescript as it is installed from package.json
- Added rush install command.


jianyexi (approved):
make sense , thanks

PR #362, opened 2022-07-21
Add SchemaTypeAndFormat rule with tests and doc

mikekistler:
This PR adds a rule to check that the type and format combination for a schema is well-defined.

### DIFF
--- a/packages/rulesets/src/spectral/functions/schema-type-and-format.ts
+++ b/packages/rulesets/src/spectral/functions/schema-type-and-format.ts
@@ -1,0 +1,104 @@
+// Check that format is valid for a schema type.
+// Valid formats are those defined in the OpenAPI spec and extensions in autorest.
+// - https://github.com/OAI/OpenAPI-Specification/blob/main/versions/2.0.md#data-types
+// - https://github.com/Azure/autorest/blob/main/packages/libs/openapi/src/v3/formats.ts
+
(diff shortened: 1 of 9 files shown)


package.json
```diff
@@ -0,0 +1,5 @@
+{
+  "dependencies": {
+    "@stoplight/types": "^13.5.0"
```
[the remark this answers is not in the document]
mikekistler: Fixed.

packages/rulesets/src/spectral/functions/schema-type-and-format.ts
```diff
@@ -0,0 +1,96 @@
...
+        path: [...path, "format"],
+      })
+    }
+  } else if (schema.properties && typeof schema.properties === "object") {
+    // eslint-disable-next-line no-restricted-syntax
+    for (const [key, value] of Object.entries(schema.properties)) {
+      errors.push(...checkSchemaTypeAndFormat(value, options, { path: [...path, "properties", key] }))
+    }
+  }
+
```
jianyexi, line 92: Seems need to check allOf as well ?

mikekistler: Yes! I fixed this in azure-api-style-guide but somehow missed including that fix here.  Fixed now.

jianyexi (approved):

PR #424, opened 2022-12-14
Add rule finding redefined common type params

bdefoy:
Add a rule to search for common type parameters by name and check if they are using a JSON reference or being redefined.


packages/rulesets/src/spectral/functions/parameter-not-using-common-types.ts
```diff
@@ -0,0 +1,39 @@
...
+    "managementGroupName",
+    "scope",
+    "tenantId",
+    "ifMatch",
+    "ifNoneMatch",
+  ])
+
+  const swagger = ctx?.documentInventory?.resolved
+
+  const allParams = parameters.concat(Object.values(swagger?.parameters ?? []))
```
jianyexi, line 27: The swagger.parameters is global parameters, seems you need to use parameters directly

bdefoy: @jianyexi I am checking for these parameter names in both the global parameters and the parameters for the paths

jianyexi (approved):

docs/parameter-not-using-common-types.md
```diff
@@ -0,0 +1,25 @@
...
+
+Some commonly used parameters are defined in the common-types directory. These parameters should be referenced instead of explicitly redefining them.
+
+## Why the rule is important
+
+To reduce duplication, maintain consistent structure in ARM specifications.
+
+## How to fix the violation
+
+Ensure any paramaters that share a name with those in common-types are declared using a JSON reference (e.g. `"$ref": ./path/to/file#/parameters/CommonParameter`).
```
jianyexi, line 25: We should point that the common types are defined in https://github.com/Azure/azure-rest-api-specs/blob/main/specification/common-types/resource-management/v4/types.json#L624

bdefoy:
@jianyexi, could you please review again merge if it looks good? I ran the command to update rules.md

PR #434, opened 2023-01-06
Validate provisioning state only if azure resource

laxmankumar12:
Perform 'Rpaas_ResourceProvisioningState' linter rule only if the value is the given resource is an azure resource (x-ms-azure-resource == true)

### DIFF
--- a/packages/rulesets/src/native/legacyRules/Rpaas_ResourceProvisioningState.ts
+++ b/packages/rulesets/src/native/legacyRules/Rpaas_ResourceProvisioningState.ts
@@ -22,14 +22,16 @@ rules.push({
     const allResources = utils.getAllResourceNames()
     for (const resource of allResources) {
       const model = utils.getResourceByName(resource)
+      const azureResource = swaggerUtil?.getProperty(model!, "x-ms-azure-resource")
       const properties = swaggerUtil?.getProperty(model!, "properties")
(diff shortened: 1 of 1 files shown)


packages/rulesets/src/native/legacyRules/Rpaas_ResourceProvisioningState.ts
```diff
@@ -22,14 +22,16 @@ rules.push({
     const allResources = utils.getAllResourceNames()
     for (const resource of allResources) {
       const model = utils.getResourceByName(resource)
+      const azureResource = swaggerUtil?.getProperty(model!, "x-ms-azure-resource")
```
laxmankumar12, line 25: I don't think it is right to read "x-ms-azure-resource" as property here. So an existing test is failing. I'll need some help on figuring this part out so that we can run this provisioningState validation only if it is an ARM resource.

jianyexi: We have a strategy to detect if a model indicates an arm resource, see https://github.com/Azure/azure-openapi-validator/blob/main/docs/resources/resource-model-validation.md, so we don't need to check the 'x-ms-azure-resource' again here, if you encounter any false alert, please comment and I will take a look

PR #449, opened 2023-02-13
Add RPC-Common-V1-05 linter rule

bdefoy:
> Error response content of long running operation MUST contain code and message properties and follow the error schema listed in the common types.


packages/rulesets/src/spectral/az-arm.ts
```diff
@@ -128,6 +128,22 @@ const ruleset: any = {
...
+      message: "{{description}}",
+      severity: "error",
+      resolved: false,
+      formats: [oas2],
+      given:
+        "$[paths,'x-ms-paths'].*.*[?(@property === 'x-ms-long-running-operation' && @ === true)]^.responses[?(@property === 'default' || @property.startsWith('5') || @property.startsWith('4'))].schema.$ref",
+      then: {
+        function: pattern,
+        functionOptions: {
+          match: ".*/common-types/resource-management/v[0-9]+/types.json#/definitions/ErrorResponse",
```
jianyexi, line 143: I think the v1/types.json is not allowed, as its schema doesn't meet the ARM RPC

bdefoy: Updated it to support v2+

docs/lro-error-response.md
```diff
@@ -0,0 +1,25 @@
...
+ARM Error
+
+## Applies to
+
+ARM OpenAPI(swagger) specs
+
+## Output Message
+
+Not using common types json ref for long running operation error response
+
```
AkhilaIlla, line 14: I thought 
This should match/be the error message that we throw

packages/rulesets/generated/spectral/az-arm.js
```diff
@@ -1665,25 +1665,6 @@ const provisioningStateSpecified = (pathItem, _opts, ctx) => {
...
-    const swagger = (_a = ctx === null || ctx === void 0 ? void 0 : ctx.documentInventory) === null || _a === void 0 ? void 0 : _a.resolved;
-    if (path === null || typeof path !== "string" || path.length === 0 || swagger === null) {
-        return [];
-    }
-    const pathRegEx = new RegExp(path.replace(scopeParameter, ".*"));
-    const otherPaths = Object.keys(swagger.paths).filter((p) => p !== path);
-    const matches = otherPaths.filter((p) => pathRegEx.test(p));
-    const errors = matches.map((match) => {
-        return {
-            message: `Path "${match}" with explicitly defined scope is a duplicate of path "${path}" that has the scope parameter.".`,
```
AkhilaIlla, line 1680: >            message: `Path "${match}" with explicitly defined scope is a duplicate of path  [](http://example.com/codeflow?start=0&length=91)

Do we checkin this file? I thought this is autogenerated & never checked it in.

packages/rulesets/src/spectral/test/lro-error-response.test.ts
```diff
@@ -0,0 +1,139 @@
...
+    },
+    definitions: {
+      ErrorResponse: {
+        description: "an error response",
+      },
+    },
+  }
+  return nonResolvingLinter.run(myOpenApiDocument).then((results) => {
+    expect(results.length).toBe(1)
+    expect(results[0].path.join(".")).toBe("paths./api/Paths.delete.responses.500.schema.$ref")
```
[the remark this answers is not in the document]
bdefoy: In V1, `code` and `message` are not under `error` (like they are [here in v2](https://github.com/Azure/azure-rest-api-specs/blob/master/specification/common-types/resource-management/v2/types.json#L322)), which goes against the RPC. I will update the message to say we require v2 and above though.

packages/rulesets/generated/spectral/az-arm.js
bdefoy: It is generated when you build the project. I'm not sure if it's necessary or not, but Roopesh [checked it in](https://github.com/Azure/azure-openapi-validator/commit/856b31b6c6d560e653b6e5fecd9e966c46563cfd#diff-2c4dbe936f4558f78f5611dc25eaba744271ce8a45191f1d92143dbd653be0cf) for at least one of his linter rules. Full history is [here](https://github.com/Azure/azure-openapi-validator/commits/main/packages/rulesets/generated/spectral/az-arm.js).

bdefoy: I think it probably does not matter in the end because the project will be rebuilt in the pipeline when it runs.

AkhilaIlla (approved):

PR #459, opened 2023-02-16
Add rule for RPC-Operations-V1-01 checking that operations API is using common types

bdefoy:


packages/rulesets/src/spectral/az-arm.ts
```diff
@@ -538,6 +538,25 @@ const ruleset: any = {
...
+    /// ARM rules for operations API
+    ///
+
+    OperationsApiSchemaUsesCommonTypes: {
+      description: "Operations API path must follow the schema provided in the common types.",
+      message: "{{description}}",
+      severity: "error",
+      resolved: false,
+      formats: [oas2],
+      given: "$[paths,'x-ms-paths'][?(@property.match(/\\/providers\\/\\w+\\.\\w+\\/operations$/i))].get.responses.200.schema.$ref",
```
AkhilaIlla, line 551: why are we checking only for get?
Will we not have a Put/Patch/Delete for Operations API?

packages/rulesets/src/spectral/az-arm.ts
```diff
@@ -538,6 +538,25 @@ const ruleset: any = {
       },
     },
 
+    ///
+    /// ARM rules for operations API
+    ///
+
+    OperationsApiSchemaUsesCommonTypes: {
```
AkhilaIlla, line 545: can we add the linter rule id "RPC-Uri-V1-10" above this

packages/rulesets/src/spectral/test/operations-api-uses-common-types.test.ts
```diff
@@ -0,0 +1,106 @@
...
+    expect(results[0].path.join(".")).toBe("paths./providers/Microsoft.MyNs/operations.get.responses.200.schema.$ref")
+    expect(results[0].message).toContain("Operations API path must follow the schema provided in the common types.")
+  })
+})
+
+test("OperationsApiSchemaUsesCommonTypes should find no errors", () => {
+  const oasDoc = {
+    swagger: "2.0",
+    paths: {
+      "/providers/Microsoft.MyNs/operations": {
```
AkhilaIlla, line 88: > "/providers/Microsoft.MyNs/operations" [](http://example.com/codeflow?start=5&length=39)

Can we add few more paths to check the regex ?

packages/rulesets/src/spectral/az-arm.ts
bdefoy: Operations API only needs the Get endpoint. Only 200 because that is the response code that would return the common type we are checking for in this rule

AkhilaIlla (approved):

packages/rulesets/src/spectral/test/operations-api-uses-common-types.test.ts
bdefoy: added some

PR #465, opened 2023-02-22
Implement rpc_delete_v1_01

AkhilaIlla:


packages/rulesets/src/spectral/functions/delete-response-codes.ts
```diff
@@ -0,0 +1,25 @@
+export const DeleteResponseCodes = (deleteOp: any, _opts: any, ctx: any) => {
+  if (deleteOp === null || typeof deleteOp !== "object") {
```
jianyexi, line 2: Seems we already implemented an rule similar to this [/home/justinxi/azure-openapi-validator/packages/rulesets/src/native/legacyRules/DeleteOperationResponses.ts](https://github.com/Azure/azure-openapi-validator/blob/main/packages/rulesets/src/native/legacyRules/DeleteOperationResponses.ts)

AkhilaIlla: @jianyexi Looks like 

'''

*run(doc, node, path) {
    // async delete is not in the scope
    if (node["202"] || node["x-ms-long-running-operation"]) {
      return
    }
    if (!node["200"] || !node["204"] || !Object.keys(node["200"]) || !Object.keys(node["204"])) {
      yield {
        message: `The delete operation is defined without a 200 or 204 error response implementation,please add it.'`,
        location: path
      }
    }
  }

'''

only covers for sync Delete and doesnt check for LRO delete

bdefoy:
Moved to #489

PR #471, opened 2023-02-28
implement Rpc-put-v1-11 

AkhilaIlla:
RPC-Put-V1-11: Response code should be 201 if resource was newly created, 200 if updated.
meaning: 201 should have the term "Created" , "create" or something that indicates creation in its description and 200 should have the term "Updated" or "Update" or "Updating" or something that indicates its an update


jianyexi (approved):

jianyexi:
this seems duplicated with last PR.

PR #472, opened 2023-03-08
Get collections must have only value and nextLink (RPC-Get-v1-09, RPC-Arg-V1-01)

bdefoy:
RPC-Get-v1-09: 
> LIST calls MUST have **ONLY** "value" and "nextLink" properties in its model.

RPC-Arg-V1-01:L
> Do not model embedded resource or reference child resource qualifications on parent resources for LIST calls. This causes inconsistent query results between portal and ARG.


konrad-jamrozik:
I would appreciate a PR description explaining this change, including what kind of testing has been conducted to confirm this change is correct.

packages/rulesets/src/spectral/az-arm.ts
```diff
@@ -211,6 +212,18 @@ const ruleset: any = {
         function: falsy,
       },
     },
+    // RPC Codes: RPC-Get-V1-09, RPC-Arg-V1-01, RPC-Get-V1-06
```
konrad-jamrozik, line 215: Can we link here to a doc elaborating on what are these codes?

bdefoy: We do not have a place where these are published yet. The descriptions are currently only available in the ARM OneNote.

konrad-jamrozik: @bdefoy could you link me to the relevant section in the ARM OneNote? I think we should clarify in the code comments and docs the canonical source of this rules is not public, and thus what is the process of cross-checking/verifying correctness of them. E.g. who to contact if there are any questions about this rule.

konrad-jamrozik:
The PR description mentions code `RPC-Get-v1-09` and `RPC-Arg-V1-01` but the added code comment also mentions `RPC-Get-V1-06`. What is this extra code about?

Also, could you help me understand how the changes made in this PR fulfill the rule `RPC-Arg-V1-01` ? I don't see it based on the code and doc changes and the PR description.

packages/rulesets/src/spectral/az-arm.ts
rkmanda: For all ARM specific rules, you need to contact ARM RPC Governance Team <armrpcgovernanceteam@microsoft.com>

rkmanda (approved):
:shipit:

PR #480, opened 2023-03-23
Add linter rule for RPC-SystemData-v1-01 and RPC-SystemData-v1-02

bdefoy:
RPC-SystemData-v1-01
> Implement System metadata for tracked resource

RPC-SystemData-v1-02
> Implement System metadata for proxy resource

This PR introduces two new linter rules, `SystemDataDefinitionsCommonTypes` and `SystemDataInPropertiesBag`, to address these ARM RPC guidelines.

`SystemDataDefinitionsCommonTypes` ensures that system metadata references are using the common-types definitions.

`SystemDataInPropertiesBag` ensures that system metadata is not defined in the properties bag, but rather as a top-level property.

Note that the common-types definitions for resources have system data already defined. These new rules are to check the specs that have their own definitions for resources and cannot switch to the common-types definitions without making breaking changes.


packages/rulesets/src/spectral/az-arm.ts
```diff
@@ -584,6 +584,42 @@ const ruleset: any = {
...
+      formats: [oas2],
+      given: "$.definitions.*.properties.[systemData,SystemData].$ref",
+      then: {
+        function: pattern,
+        functionOptions: {
+          match: ".*/common-types/resource-management/v\\d+/types.json#/definitions/systemData",
+        },
+      },
+    },
+    // RPC Code: RPC-SystemData-V1-01
```
rkmanda, line 609: add new line

packages/rulesets/src/spectral/az-arm.ts
```diff
@@ -584,6 +584,42 @@ const ruleset: any = {
       },
     },
 
+    ///
+    /// ARM RPC rules for SystemData
+    ///
+    // RPC Code: RPC-SystemData-V1-01
+    // Note that this rule is normally already covered when the common types Resource is used.
+    // Here we are still checking for usage of common types system data for the scenarios where
+    // RP teams have already defined Resource and are unable to switch to the common types resource
```
rkmanda, line 593: > defined Resource and are unable to switch to the common types resource [](http://example.com/codeflow?start=28&length=71)

Shouldnt we be checking the actual definition in that case? I see the main thing being checked if the common-type is being used here. Is my understanding correct?

rkmanda (changes requested):
:clock1:

packages/rulesets/src/spectral/az-arm.ts
bdefoy: We previously discussed that I should also create common types for tracked and proxy resources without the system data so that teams without system data can still use common types. I started that [here](https://github.com/Azure/azure-rest-api-specs/pull/23341/files) but still need to finish that PR. After that, we decided that we can just check that common types are used here.

bdefoy: Yes, that is correct. The idea here is that by ensuring the common-types definition for systemData is used, you are ensuring that the body is correct.

packages/rulesets/src/spectral/test/system-data-in-properties-bag.test.ts
```diff
@@ -0,0 +1,62 @@
...
+let nonResolvingLinter: Spectral
+const RULE = "SystemDataInPropertiesBag"
+const ERROR_MESSAGE = "System Data must be defined as a top-level property, not in the properties bag."
+
+beforeAll(async () => {
+  nonResolvingLinter = await linterForRule(RULE, true)
+  return nonResolvingLinter
+})
+
+test(`${RULE} should find errors`, async () => {
```
rkmanda, line 13: >should [](http://example.com/codeflow?start=14&length=6)

Lets add another test for completeness where systemData is defined in a model referenced from properties.properties instead of inlining it

docs/system-data-definitions-common-types.md
```diff
@@ -0,0 +1,22 @@
...
+- RPC-SystemData-V1-01
+- RPC-SystemData-V1-02
+
+## Description
+
+Validates that system data references utilize the common types.
+
+## How to fix the violation
+
+Ensure that references to system data use the common types definition for system data.
```
konrad-jamrozik, line 22: Curious: why we didn't provide valid and invalid examples?

docs/system-data-in-properties-bag.md
```diff
@@ -0,0 +1,48 @@
...
+  }
+}
+```
+
+### Invalid/Bad Example
+
+```json
+"Resource": {
+  "properties": {
+    "properties": {
```
konrad-jamrozik, line 41: Is it common to have nested `properties/properties` ? If not, maybe let's use names that are more realistic?

packages/rulesets/src/spectral/az-arm.ts
```diff
@@ -707,6 +707,44 @@ const ruleset: any = {
...
     },
 
+    ///
+    /// ARM RPC rules for SystemData
+    ///
+
+    // RPC Code: RPC-SystemData-V1-01 and RPC-SystemData-V1-02
+    // Note that this rule is normally already covered when the common types Resource is used.
+    // Here we are still checking for usage of common types system data for the scenarios where
+    // RP teams have already defined Resource and are unable to switch to the common types resource
```
konrad-jamrozik, line 717:
```suggestion
    // RP teams have already defined Resource and are unable to switch to the common types Resource
```

or `resource`, but then let's make the capitalization consistent

packages/rulesets/src/spectral/az-arm.ts
```diff
@@ -707,6 +707,44 @@ const ruleset: any = {
       },
     },
 
+    ///
+    /// ARM RPC rules for SystemData
+    ///
+
+    // RPC Code: RPC-SystemData-V1-01 and RPC-SystemData-V1-02
+    // Note that this rule is normally already covered when the common types Resource is used.
```
konrad-jamrozik, line 715: What is already covering it? Another rule? Can we reference it in this comment?

packages/rulesets/generated/spectral/az-arm.js
```diff
@@ -3039,6 +3039,31 @@ const ruleset = {
                 function: noDuplicatePathsForScopeParameter,
             },
         },
+        SystemDataDefinitionsCommonTypes: {
+            description: "Must use the schema provided in the common types for SystemData.",
```
konrad-jamrozik, line 3043: Is there a reason we are not using the same description as in `rules.md` ? Something like:

> System data references must utilize common types.

In general, I am unsure why we would want for the description in `rules.md` be mismatched with the error description.

docs/system-data-in-properties-bag.md
konrad-jamrozik: OK I found below

`given: ["$.definitions.*.properties.properties.systemData", "$.definitions.*.properties.properties.SystemData"],`

so I guess `properties.properties` is the standard.

packages/rulesets/src/spectral/az-arm.ts
```diff
@@ -707,6 +707,44 @@ const ruleset: any = {
...
+        functionOptions: {
+          match: ".*/common-types/resource-management/v\\d+/types.json#/definitions/systemData",
+        },
+      },
+    },
+
+    // RPC Code: RPC-SystemData-V1-01 and RPC-SystemData-V1-02
+    // Ensure systemData is not in the properties bag
+    SystemDataInPropertiesBag: {
+      description: "System Data must be defined as a top-level property, not in the properties bag.",
```
konrad-jamrozik, line 737:
```suggestion
      description: "SystemData must be defined as a top-level property, not in the properties bag.",
```

docs/system-data-in-properties-bag.md
bdefoy: Yes, the first level of `properties` is part of the OpenAPI specification syntax and the second level is describing the Resource to have a field called `properties`.

docs/system-data-definitions-common-types.md
bdefoy: We don't have a set of required sections for the docs as far as I know. I can add them here though.

packages/rulesets/src/spectral/az-arm.ts
bdefoy: When the common-types resource definition is used there won't be any system data references because the common-types resource definition has system data. So, this rule is only applicable for specs that are not using the common-types resource definition. Updating the comment to clarify this.

packages/rulesets/generated/spectral/az-arm.js
bdefoy: Changed them to that

packages/rulesets/src/spectral/az-arm.ts
bdefoy: Changed it to "System data" to match the other rule's description.

rkmanda (approved):
:shipit:

PR #485, opened 2023-03-29
Add rule to flag explicitly-defined error code responses (not using default response)

bdefoy:


packages/rulesets/src/spectral/functions/no-error-code-responses.ts
```diff
@@ -0,0 +1,26 @@
...
+
+  // no error if the response code matches any of the allowed response codes
+  if (ALLOWED_RESPONSE_CODES.some((allowedCode) => responseCode === allowedCode)) {
+    return []
+  }
+
+  // response code did not match any of the allowed codes, so error
+  return [
+    {
+      message: "",
```
rkmanda, line 22: >message: "", [](http://example.com/codeflow?start=6&length=12)

what error msg would be shown?

packages/rulesets/src/spectral/az-arm.ts
```diff
@@ -620,6 +621,17 @@ const ruleset: any = {
         field: "type",
       },
     },
+    NoErrorCodeResponses: {
+      description: "Responses must not have error codes. All errors must be surfaced using `default`.",
```
rkmanda, line 625: >Responses must not have error codes [](http://example.com/codeflow?start=20&length=35)

Invalid status code specified. Please refer the documentation for the allowed set.

rkmanda (changes requested):
:clock1:

rkmanda (approved):
:shipit:

PR #488, opened 2023-04-03
Flag error if provisioning state property is not set as read Only

AkhilaIlla:


packages/rulesets/src/spectral/az-arm.ts
```diff
@@ -608,6 +609,17 @@ const ruleset: any = {
     /// ARM rules without an RPC code
     ///
 
+    ProvisioningStateMustBeReadOnly: {
+      description: "This is a rule introduced to validate if provisioningState property is set to readOnly or not.",
+      message: "{{error}}",
+      severity: "error",
+      resolved: true,
+      formats: [oas2],
+      given: ["$[paths,'x-ms-paths'].*.*.responses[*].schema"],
```
bdefoy, line 618: I think this is equivalent:
```suggestion
      given: ["$[paths,'x-ms-paths'].*.*.responses.*.schema"],
```

packages/rulesets/src/spectral/functions/provisioning-state-must-be-read-only.ts
```diff
@@ -0,0 +1,26 @@
...
+export const provisioningStateMustBeReadOnly = (schema: any, _opts: any, ctx: any) => {
+  if (schema === null || typeof schema !== "object") {
+    return []
+  }
+
+  const path = ctx.path || []
+  const errors = []
+
+  const allProperties = getProperties(schema)
+  const provisioningStateProperty = getProperty(allProperties?.properties, "provisioningState")
```
bdefoy, line 12: Looking at the utils functions, it looks like you can just use `getProperty` directly, and do not need to use `getAllProperties` in this case.
```suggestion
  const provisioningStateProperty = getProperty(schema, "provisioningState")
```

packages/rulesets/src/spectral/az-arm.ts
AkhilaIlla: yeah & it'll be consistent as well!

packages/rulesets/src/spectral/functions/provisioning-state-must-be-read-only.ts
AkhilaIlla: I used the same in one of my previous rules & it started false negatives as ProvisioningState property is within the properties bucket in properties, so we do one level deep and then get the property

bdefoy (approved):

PR #489, opened 2023-04-04
[Formerly #465] Implement RPC-Delete-V1-01

bdefoy:
Synchronous delete operations must have 200, 204, and default responses. They must not have any other responses.
Long-running (LRO) delete operations must have 202, 204, and default responses. They must not have any other responses.

Moved from #465


packages/rulesets/src/spectral/functions/delete-response-codes.ts
```diff
@@ -0,0 +1,29 @@
...
+  }
+  const path = ctx.path
+  const errors = []
+
+  if (!deleteOp?.responses) {
+    return []
+  }
+
+  if (deleteOp["x-ms-long-running-operation"] && deleteOp["x-ms-long-running-operation"] === true) {
+    if (!deleteOp.responses["202"] || !deleteOp.responses["204"]) {
```
rkmanda, line 13: >deleteOp.responses["204"] [](http://example.com/codeflow?start=39&length=25)

lets also check that no other status code is specified

packages/rulesets/src/spectral/functions/delete-response-codes.ts
```diff
@@ -0,0 +1,29 @@
...
+      errors.push({
+        message: "Long-running (LRO) delete operations must have responses with 202 and 204 return codes.",
+        path: path,
+      })
+    }
+  } else {
+    if (!deleteOp.responses["200"] || !deleteOp.responses["204"]) {
```
rkmanda, line 20: >!deleteOp.responses["204"]) { [](http://example.com/codeflow?start=38&length=29)

lets also check that no other status code is specified

rkmanda (changes requested):
:clock1:

packages/rulesets/src/spectral/functions/delete-response-codes.ts
bdefoy: @rkmanda should I have the rule ensure `202`, `204`, and `default` are all specified and no others for LROs?

bdefoy:
Might be able to merge this into LongRunningResponseStatusCode #491

packages/rulesets/src/spectral/functions/delete-response-codes.ts
bdefoy, line 1: Might be able to merge this into LongRunningResponseStatusCode #491

bdefoy: This rule ensures that the response codes are there, while the other rule ensures that at least one success code is there.

packages/rulesets/src/spectral/functions/delete-response-codes.ts
```diff
@@ -0,0 +1,37 @@
...
+  if (!deleteOp?.responses) {
+    return []
+  }
+
+  const responses = Object.keys(deleteOp?.responses ?? {})
+
+  if (deleteOp["x-ms-long-running-operation"] === true) {
+    if (responses.length !== LR_DELETE_RESPONSES.length || !LR_DELETE_RESPONSES.every((value) => responses.includes(value))) {
+      errors.push({
+        message: "Long-running (LRO) delete operations must have responses with 202 and 204 return codes, and no other response codes.",
```
konrad-jamrozik, line 23: and what about default? The doc also mentions default.

docs/delete-response-codes.md
```diff
@@ -0,0 +1,27 @@
...
+
+ARM OpenAPI (swagger) specs
+
+## Related ARM Guideline Code
+
+- RPC-Delete-V1-01
+
+## Output Message
+
+Synchronous delete operations must have 200, 204, and default responses. They must not have any other responses.
```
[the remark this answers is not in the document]
konrad-jamrozik: By reading the code I think you mean both.

packages/rulesets/src/spectral/test/delete-response-codes.test.ts
```diff
@@ -0,0 +1,588 @@
...
+const SYNC_ERROR = "Synchronous delete operations must have responses with 200 and 204 return codes, and no other response codes."
+
+let linter: Spectral
+
+beforeAll(async () => {
+  linter = await linterForRule("DeleteResponseCodes")
+  return linter
+})
+
+test("DeleteResponseCodes should find errors for sync delete with only 200", () => {
```
konrad-jamrozik, line 14: I think we could also add a test in which it returns all the required codes, abut also some extra forbidden codes.

packages/rulesets/src/spectral/test/delete-response-codes.test.ts
```diff
@@ -0,0 +1,588 @@
...
+    },
+  }
+  return linter.run(myOpenApiDocument).then((results) => {
+    expect(results.length).toBe(1)
+    expect(results[0].path.join(".")).toBe("paths./foo.delete")
+    expect(results[0].message).toContain(LR_ERROR)
+  })
+})
+
+test("DeleteResponseCodes should find no errors for lro delete", () => {
```
konrad-jamrozik, line 513:
```suggestion
test("DeleteResponseCodes should find no errors for lro delete when all required codes are provided", () => {
```

I suggest similar fix to other relevant tests.

docs/delete-response-codes.md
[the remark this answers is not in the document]
bdefoy: It's like a catch-all for any of the codes that are not specified. It's used to have the same error response body across all error response codes. https://swagger.io/docs/specification/describing-responses/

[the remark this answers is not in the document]
bdefoy: Yes, both. The rule ensures that the spec has all those responses defined, which covers the first point. We can't ensure the second point with a linter rule (that requires some sort of live validation) but that is the expectation.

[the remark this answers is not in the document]
konrad-jamrozik: Sorry, I have another basic question: why we cannot validate with linter the second rule, but can the first? I mean, we could claim in the swagger spec that we return 200, 204 and so on, but in reality implement it in such a way that it only ever returns 200. Then we will break the first rule, and we won't be able to validate even with live validation, as it cannot can only tell us what it saw, but not what it will never see.

konrad-jamrozik (approved):

mikekistler (approved):
Looks good! 👍

rkmanda (approved):

PR #493, opened 2023-05-01
Affix `-rulesets` package dependency to `1.0.1` and update `-core` package to `1.0.1`

konrad-jamrozik:
In preparation to deploy [`@microsoft.azure/openapi-validator`](https://www.npmjs.com/package/@microsoft.azure/openapi-validator?activeTab=versions) `v2.0.1`, affix ruleset used to `1.0.1` and update [`@microsoft.azure/openapi-validator-core`](https://www.npmjs.com/package/@microsoft.azure/openapi-validator-core?activeTab=versions) from `1.0.0` to `1.0.1`.

This PR is step `1.1` from the [deployment plan](https://github.com/Azure/azure-sdk-tools/issues/6071#issuecomment-1529570291) of:
- https://github.com/Azure/azure-sdk-tools/issues/6071

### DIFF
--- a/packages/azure-openapi-validator/core/package.json
+++ b/packages/azure-openapi-validator/core/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@microsoft.azure/openapi-validator-core",
-  "version": "1.0.0",
+  "version": "1.0.1",
   "description": "Azure OpenAPI Validator",
   "main": "dist/index.js",
   "scripts": {
--- a/changelog.md
+++ b/changelog.md
@@ -4,7 +4,7 @@
 
 In preparation for deployment of pending ruleset changes to production,
 affixed package `@microsoft.azure/openapi-validator-rulesets` version used by `@microsoft.azure/openapi-validator`
-to be exactly `1.0.1`.
+to be exactly `1.0.1`. Also updated `@microsoft.azure/openapi-validator-core` to be `1.0.1`.
 
 For details, see:  
   Deploy pending azure-openapi-validator (LintDiff) changes from staging to prod #6071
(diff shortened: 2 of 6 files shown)


packages/azure-openapi-validator/core/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@microsoft.azure/openapi-validator-core",
-  "version": "1.0.0",
+  "version": "1.0.1",
```
konrad-jamrozik, line 3: FYI: If I would want to play this extra-safe, I would change the azure-openapi-validator `package.json` dependency from `"@microsoft.azure/openapi-validator-core": "~1.0.0",` to `"@microsoft.azure/openapi-validator-core": "1.0.0",` and then update it later to `1.0.1`. But I deem the risk acceptable here. If things break, I will deploy core `1.0.2` that reverts to `1.0.0`. There appears to be only one small commit since Nov 2022 to core that needs to be picked up: https://github.com/Azure/azure-openapi-validator/commits/main/packages/azure-openapi-validator/core

weshaggard (approved):

konrad-jamrozik:
Looks like this PR together with the LINT_VERSION update possibly broke LintDiff:

https://dev.azure.com/azure-sdk/internal/_build/results?buildId=2745334&view=logs&j=0574a2a6-2d0a-5ec6-40e4-4c6e2f70bea2&t=80c3e782-49f0-5d1c-70dd-cbee57bdd0c7
from 
https://github.com/Azure/azure-rest-api-specs/pull/23800/checks?check_run_id=13200514347

Need to dig into this. Maybe because of some missing `rush update` or similar commands. Need to dig into this. @weshaggard FYI.

PR #499, opened 2023-05-04
checks the properties of type:object has no definition model

tejaswiMinnu:


docs/properties-type-object-no-definition.md
```diff
@@ -0,0 +1,41 @@
+# PropertiesTypeObjectNoDefinition
+
+## Category
+
```
rkmanda, line 4: also include 

## Related ARM Guideline Code

- RPC-Put-V1-22

packages/rulesets/src/spectral/az-arm.ts
```diff
@@ -224,6 +225,20 @@ const ruleset: any = {
...
 
+    // RPC Code: RPC-Policy-V1-03
+    PropertiesTypeObjectNoDefinition: {
+      description:
+        "If Properties with type:object dont have a reference model defined, then the allowed types can only be primitive data types.",
+      severity: "error",
+      message: "{{description}}",
+      resolved: true,
+      formats: [oas2],
+      given: "$..[?(@property === 'type' && @ === 'object' && @.properties === undefined)]^",
```
rkmanda, line 236: > @.properties === undefined) [](http://example.com/codeflow?start=61&length=28)

you need to cover the case where the .properties exists but is empty as well

packages/rulesets/src/spectral/functions/properties-type-object-no-definition.ts
```diff
@@ -0,0 +1,21 @@
+// If a property is defined with `'type': 'object'`, it must also define an object body.
+// RPC Code: RPC-Policy-V1-03
+
+export const propertiesTypeObjectNoDefinition: any = (definitionObject: any, _opts: any, ctx: any) => {
+  const values = Object.values(definitionObject)
+
+  for (let value of values) {
+    if (values.length === 1) {
+      if (value === "object") {
```
rkmanda, line 9: This logic is unclear to me. Lets discuss.

rkmanda (changes requested):
:clock1:

PR #500, opened 2023-05-04
Promote linter warnings to errors

rkmanda:
Promote a couple of linter rules which are currently marked as warnings to errors

LroExtension
LroStatusCodesReturnTypeSchema
DescriptiveDescriptionRequired
ParameterDescriptionRequired

### DIFF
--- a/docs/lro-status-codes-return-type-schema.md
+++ b/docs/lro-status-codes-return-type-schema.md
@@ -2,7 +2,7 @@
 
 ## Category
 
-SDK Warning
+SDK Error
 
 ## Applies to
 
--- a/packages/rulesets/src/spectral/az-common.ts
+++ b/packages/rulesets/src/spectral/az-common.ts
@@ -21,6 +21,7 @@ import { operationIdNounVerb } from "./functions/operation-id-noun-verb"
 import paramLocation from "./functions/parameter-location"
 import { patchInOperationName } from "./functions/patch-in-operation-name"
 import { putInOperationName } from "./functions/put-in-operation-name"
+import { putRequestResponseScheme } from "./functions/put-request-response-scheme"
 import { requiredReadOnlyProperties } from "./functions/required-read-only-properties"
 import checkSchemaFormat from "./functions/schema-format"
 import checkSummaryAndDescription from "./functions/summary-description-must-not-be-same"
@@ -96,7 +97,7 @@ const ruleset: any = {
     LroExtension: {
       description: "Operations with a 202 response should specify `x-ms-long-running-operation: true`.",
       message: "Operations with a 202 response should specify `x-ms-long-running-operation: true`.",
-      severity: "warn",
+      severity: "error",
       formats: [oas2],
       given: "$.paths[*][*].responses[?(@property == '202')]^^",
       then: {
@@ -107,7 +108,7 @@ const ruleset: any = {
     LroStatusCodesReturnTypeSchema: {
       description: "The '200'/'201' responses of the long running operation must have a schema definition.",
       message: "{{error}}",
-      severity: "warn",
+      severity: "error",
       resolved: true,
       formats: [oas2],
       given: ["$[paths,'x-ms-paths'].*[put][?(@property === 'x-ms-long-running-operation' && @ === true)]^"],
@@ -204,6 +205,17 @@ const ruleset: any = {
         function: deleteInOperationName,
       },
     },
+    PutRequestResponseScheme: {
+      description: "The request & response('200' and '201') schema of the PUT operation must be same.",
+      message: "{{error}}",
+      severity: "error",
+      resolved: true,
+      formats: [oas2],
+      given: ["$[paths,'x-ms-paths'].*[put][responses][?(@property === '200' || @property === '201')]^^"],
+      then: {
+        function: putRequestResponseScheme,
+      },
+    },
     RequiredReadOnlyProperties: {
       description:
         "A model property cannot be both `readOnly` and `required`. A `readOnly` property is something that the server sets when returning the model object while `required` is a property to be set when sending it as a part of the request body.",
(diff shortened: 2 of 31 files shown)


packages/rulesets/src/spectral/az-common.ts
```diff
@@ -484,7 +496,7 @@ const ruleset: any = {
       severity: "warn",
       resolved: false,
       formats: [oas2],
-      given: ["$..[?(@object() && @.properties)].properties[?(@object() && @.properties)].properties^"],
+      given: ["$..[?(@object() && @.properties)][?(@object() && @.properties)].properties"],
```
rkmanda, line 499: >properties [](http://example.com/codeflow?start=36&length=10)

check why this is showing up as a change?

packages/rulesets/src/spectral/az-common.ts
```diff
@@ -204,6 +205,17 @@ const ruleset: any = {
...
     },
+    PutRequestResponseScheme: {
+      description: "The request & response('200' and '201') schema of the PUT operation must be same.",
+      message: "{{error}}",
+      severity: "error",
+      resolved: true,
+      formats: [oas2],
+      given: ["$[paths,'x-ms-paths'].*[put][responses][?(@property === '200' || @property === '201')]^^"],
+      then: {
+        function: putRequestResponseScheme,
```
rkmanda, line 216: check why this whole block is showing up as a change.

packages/rulesets/src/spectral/az-common.ts
```diff
@@ -21,6 +21,7 @@ import { operationIdNounVerb } from "./functions/operation-id-noun-verb"
 import paramLocation from "./functions/parameter-location"
 import { patchInOperationName } from "./functions/patch-in-operation-name"
 import { putInOperationName } from "./functions/put-in-operation-name"
+import { putRequestResponseScheme } from "./functions/put-request-response-scheme"
```
rkmanda, line 24: >"./ [](http://example.com/codeflow?start=41&length=3)

not introduced by me. check why this shows up in the diff

packages/rulesets/src/spectral/az-common.ts
```diff
@@ -249,7 +249,7 @@ const ruleset: any = {
       },
     },
     LongRunningOperationsOptionsValidator: {
-      description: 'A LRO Post operation with return schema must have "x-ms-long-running-operation-options" extension enabled.',
+      description: 'A LRO Post operation with return schema should have "x-ms-long-running-operation-options" extension enabled.',
```
ravimeda, line 252: Minor - I would use `must` for action to take to address an error.

ravimeda (approved):

ravimeda:
Curious - What previously was a warning would now become an error, would that break someone/something?

bdefoy (changes requested):
- Since these are all common between data plane and control plane, have we confirmed that the data plane reviewers also want these to be errors? 
- Need to update the respective .md file for each rule to say error instead of warning
- After updating the .md files, need to run `rush regen-ruleindex` to regenerate rules.md
- Need to run `rush change` to add a changelog/fix the build error

rkmanda:
Yes I gave a headsup to Mike from data plane and have given him a link to review this PR

---
In reply to: [1535484568](https://github.com/Azure/azure-openapi-validator/pull/500#issuecomment-1535484568) [](http://example.com/codeflow?ancestors=1535484568,1535484568)

rkmanda:
Yes I gave a headsup to Mike from data plane and have given him a link to review this PR. Didnt realize the md files needed an explicit update. Will do. Thanks!

---
In reply to: [1414001723](https://github.com/Azure/azure-openapi-validator/pull/500#issuecomment-1414001723) [](http://example.com/codeflow?ancestors=1414001723,1414001723)

packages/rulesets/src/spectral/az-common.ts
rkmanda: This is a warning though, not an error as indicated by severity: "warn",.

packages/rulesets/src/spectral/az-common.ts
```diff
@@ -96,7 +96,7 @@ const ruleset: any = {
     LroExtension: {
       description: "Operations with a 202 response should specify `x-ms-long-running-operation: true`.",
```
rkmanda, line 97: >should [](http://example.com/codeflow?start=51&length=6)

change to must

bdefoy:
> * Since these are all common between data plane and control plane, have we confirmed that the data plane reviewers also want these to be errors?
> * Need to update the respective .md file for each rule to say error instead of warning
> * After updating the .md files, need to run `rush regen-ruleindex` to regenerate rules.md
> * Need to run `rush change` to add a changelog/fix the build error
> 
> In reply to: [1414001723](https://github.com/Azure/azure-openapi-validator/pull/500#issuecomment-1414001723)

Yeah, I wonder how difficult it would be to update error/warning in the docs based on changes in ruleset automatically.

bdefoy (approved):

docs/lro-status-codes-return-type-schema.md
```diff
@@ -2,7 +2,7 @@
 
 ## Category
 
-SDK Warning
+SDK Error
```
mikekistler, line 5: Question on this one. We now have support in autorest, and more recently in aaz-dev-tools (CLI generator), for the `final-state-schema` property in `x-ms-long-running-operation-options`:

https://github.com/Azure/autorest/tree/main/docs/extensions#x-ms-long-running-operation

If a service uses this new option, it does not need to "lie" about returning a 200 or 201 response from a long-running operation. So do we really want to make this case an error?

rkmanda: The rule is scoped to Put operations only. And it states that if a 200 or a 201 are specified then they need a schema to be defined

mikekistler: Okay. It would be nice if the rule description actually said this. But I suppose that can be fixed later.

mikekistler (approved):
Looks good. 👍

PR #504, opened 2023-05-08
Revert "The payload must not contain any additionalProperties field except for tags"

konrad-jamrozik:
Reverts Azure/azure-openapi-validator#494

Reason: this PR breaks local execution.

How to reproduce:
- clone branch of this run: https://dev.azure.com/azure-sdk/internal/_build/results?buildId=2753856&view=logs&j=688669d0-441c-57c3-cf6d-f89a22ccfa5d&t=b91b1e88-b042-5e18-36d8-34e4fb3a9b3b&l=81
- run:
  ```
  autorest --v3 --spectral --azure-validator
  --use=./packages/azure-openapi-validator/autorest
  --tag=package-2022-09-preview
  ../azure-rest-api-specs-pr/specification/containerservice/resource- 
  manager/Microsoft.ContainerService/fleet/readme.md
  ```

Observe:
```
> autorest --v3 --spectral --azure-validator --use=./packages/azure-openapi-validator/autorest --tag=package-2022-09-preview ../azure-rest-api-specs-pr/specification/containerservice/resource-manager/Microsoft.ContainerService/fleet/readme.md
AutoRest code generation utility [cli version: 3.6.3; node: v19.1.0]
(C) 2018 Microsoft Corporation.
https://aka.ms/autorest
info    | AutoRest core version selected from configuration: ^3.2.0.
info    |    Loading AutoRest core      'C:\users\username\.autorest\@autorestcore@3.9.5\nodemodules\@autorest\core\dist' (3.9.5)
info    |    Loading local AutoRest extension '@microsoft.azure/openapi-validator' (C:\users\username\repos\azure-openapi-validator\packages\azure-openapi-validator\autorest)
warning | IgnoredPropertyNextToRef | Semantic violation: Sibling values alongside $ref will be ignored. See https://github.com/Azure/autorest/blob/main/docs/openapi/howto/$ref-siblings.md for allowed values (components > schemas > Resource > properties > systemData)
  keys: [ 'type' ]

C:\users\username\.autorest\@autorest_core@3.9.5\node_modules\@autorest\core\dist\vendors-common_temp_node_modules_pnpm_azure_logger_1_0_3_node_modules_azure_logger_dist-esm_s-fe57a1.js:110268
                    throw new Error('Header must provide a Content-Length property.');
                          ^
Error: Header must provide a Content-Length property.
[...]

PR #510, opened 2023-05-10
avoid additional properties except for user defined tags

tejaswiMinnu:
Added a rule to check if any "additionalProperties" exists in the swagger.
If there are any "additionalProperties" having parent defined as "tags" ignore this path in this rule and consider this as exceptional case by allowing this path.


konrad-jamrozik:
@tejaswiMinnu as discussed on Teams, I would appreciate if you could confirm this PR does not introduce the issue described here:
- https://github.com/Azure/azure-openapi-validator/pull/504

konrad-jamrozik (changes requested):
I am requesting for [this ask](https://github.com/Azure/azure-openapi-validator/pull/510#issuecomment-1542636548) to be fulfilled before the PR is merged.

konrad-jamrozik (approved):
@tejaswiMinnu confirmed that the issue is gone now.

docs/avoid-additional-properties.md
```diff
@@ -0,0 +1,45 @@
...
+
+ARM OpenAPI(swagger) specs
+
+## Related ARM Guideline Code
+
+- RPC-Policy-V1-05
+
+## Output Message
+
+Definitions must not have properties named "additionalProperties" exept for user defined tags.
```
rkmanda, line 17: >exept [](http://example.com/codeflow?start=66&length=5)

except

packages/rulesets/src/spectral/test/avoid-additional-properties.test.ts
```diff
@@ -0,0 +1,172 @@
...
+import linterForRule from "./utils"
+
+let linter: Spectral
+const errorMessage = "The use of additionalProperties is not allowed except for user defined tags on tracked resources."
+beforeAll(async () => {
+  linter = await linterForRule("AvoidAdditionalProperties")
+  return linter
+})
+
+test("AvoidAdditionalProperties should find errors", () => {
```
rkmanda, line 11: >AvoidAdditionalProperties [](http://example.com/codeflow?start=6&length=25)

Can u add a test for when tags is not directly under properties and check what the behavior is?

rkmanda (changes requested):
:clock1:

packages/rulesets/src/spectral/test/avoid-additional-properties.test.ts
rkmanda: Is there any other rule that would catch the existence of "tags" other than under properties?

tejaswiMinnu: Added the test case

packages/rulesets/src/spectral/test/avoid-additional-properties.test.ts
```diff
@@ -0,0 +1,224 @@
...
+      info: {
+        type: "object",
+        tags: {
+          type: "object",
+          additionalProperties: {
+            type: "object",
+            params: {
+              type: "boolean",
+            },
+          },
```
rkmanda, line 207: This is not a good representation of swagger. In an actual swagger file, info will have properties underneath it and then the tags would show up under properties. Can u use an example that closely represents an RP swagger?

packages/rulesets/src/spectral/test/avoid-additional-properties.test.ts
rkmanda: tx for adding the test. I have a comment on improving the test case, pl take a look.

packages/rulesets/src/spectral/test/avoid-additional-properties.test.ts
tejaswiMinnu: I added tags under properties now to represent swagger, but either case the given expression will check for "additionalProperties" if it has parent as "tags" irrespective of where it's defined.

rkmanda: yes I understand the logic. But lets make sure to stick to a proper swagger def in all our tests

rkmanda (approved):
:shipit:

PR #513, opened 2023-05-12
Fix RPC_Patch_V1_02

rkmanda:
The rule RPC-Patch_V1_02 is to check if the top-level properties like name, type, location are not mentioned in the request body of a Patch operation. The exception to this rule is that it is ok for these properties to be specified as long as they are appropriately mentioned either as readOnly or immutable properties.

The earlier implementation of this rule did not take this exception into account and the current PR addresses this gap. The reason it is ok for readOnly and immutable properties to be specified in the payload is because the service is expected to treat these as pass thru as long as the values captured in the service are the same as the values being passed in with the payload.

### DIFF
--- a/common/changes/@microsoft.azure/openapi-validator-rulesets/rmanda-fix_rpc_patch_v1_02_2023-05-12-05-42.json
+++ b/common/changes/@microsoft.azure/openapi-validator-rulesets/rmanda-fix_rpc_patch_v1_02_2023-05-12-05-42.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@microsoft.azure/openapi-validator-rulesets",
+      "comment": "fixing rpc_patch_v1_02",
+      "type": "patch"
+    }
+  ],
+  "packageName": "@microsoft.azure/openapi-validator-rulesets"
+}
--- a/docs/un-supported-patch-properties.md
+++ b/docs/un-supported-patch-properties.md
@@ -14,7 +14,7 @@ ARM OpenAPI(swagger) specs
 
 ## Output Message
 
-The patch operation body parameter schema should not contains property {propertyName}.
+The patch operation body parameter schema should not contain top level "id", "name", "type", "location" as writable properties. This is because these properties are expected to be either readOnly or immutable.
 
 ## Description
 
@@ -30,4 +30,4 @@ July 07, 2022
 
 ## How to fix the violation
 
-Considering removing the name, location, or type in the patch request body.
+Consider either removing the top level properties - id, name and type, from the patch request body or mark them as readOnly. For the top level location property for tracked resources, consider either removing it from the request body of the Patch operation or mark it as immutable using the x-ms-mutability property and values as "create" and "read".
--- a/docs/un-supported-patch-properties.md
+++ b/docs/un-supported-patch-properties.md
@@ -14,7 +14,7 @@ ARM OpenAPI(swagger) specs
 
 ## Output Message
 
-The patch operation body parameter schema should not contain top level "id", "name", "type", "location" as writable properties. This is because these properties are expected to be either readOnly or immutable.
+The patch operation body parameter schema should not contain top level "id", "name", "type", "location" as writable properties. This is because these properties are not patchable. To fix this, either remove the offending properties from the request payload of the Patch operation, or mark them as readOnly or immutable.
 
 ## Description
 
--- a/docs/un-supported-patch-properties.md
+++ b/docs/un-supported-patch-properties.md
@@ -30,4 +30,4 @@ July 07, 2022
 
 ## How to fix the violation
 
-Consider either removing the top level properties - id, name and type, from the patch request body or mark them as readOnly. For the top level location property for tracked resources, consider either removing it from the request body of the Patch operation or mark it as immutable using the x-ms-mutability property and values as "create" and "read".
+Consider either removing the top-level properties - "id", "name" and "type", from the patch request body parameter schema, or mark them as readOnly. For the top-level "location" property (that is specified for tracked resources), consider either removing it from the request body of the Patch operation or mark it as immutable using the x-ms-mutability property and values as "create" and "read".
--- a/packages/rulesets/src/spectral/functions/validate-patch-body-param-properties.ts
+++ b/packages/rulesets/src/spectral/functions/validate-patch-body-param-properties.ts
@@ -56,26 +56,23 @@ export const validatePatchBodyParamProperties = createRulesetFunction<unknown, O
         _opts.shouldNot.forEach((p: string) => {
           const property = getProperties(bodyParameter)?.[p];
           if (property) {
-            let isPropertyBoolean = false;
+            let isPropertyReadOnly = false;
             let isPropertyImmutable = false;
 
-            if (property["readOnly"] && property["readOnly"] === true)
-            {
-              isPropertyBoolean = true;
+            if (property["readOnly"] && property["readOnly"] === true) {
+              isPropertyReadOnly = true;
             }
 
             if (property["x-ms-mutability"] && Array.isArray(property["x-ms-mutability"])) {
-                const schemeArray = property["x-ms-mutability"];
-                if (!schemeArray.includes("update"))
-                {
-                    isPropertyImmutable = true;
-                }
+              const schemaArray = property["x-ms-mutability"];
+              if (!schemaArray.includes("update")) {
+                isPropertyImmutable = true;
+              }
             }
 
-            if (!isPropertyBoolean && !isPropertyImmutable)
-            {
+            if (!(isPropertyReadOnly || isPropertyImmutable)) {
               errors.push({
-                message: `The patch operation body parameter schema should not contain property ${p}.`,
+                message: `Mark the top level property - ${p} specified in the patch operation body as readOnly or immutable. These properties are not patchable.`,
                 path: [...path, "parameters", index],
               })
             }
--- a/packages/rulesets/src/spectral/functions/validate-patch-body-param-properties.ts
+++ b/packages/rulesets/src/spectral/functions/validate-patch-body-param-properties.ts
@@ -46,7 +46,7 @@ export const validatePatchBodyParamProperties = createRulesetFunction<unknown, O
         _opts.should.forEach((p: string) => {
           if (!getProperties(bodyParameter)?.[p] && getProperties(responseSchema)?.[p]) {
             errors.push({
-              message: `The patch operation body parameter schema should contains property '${p}'.`,
+              message: `The patch operation body parameter schema should contain property '${p}'.`,
               path: [...path, "parameters", index],
             })
           }
@@ -56,8 +56,8 @@ export const validatePatchBodyParamProperties = createRulesetFunction<unknown, O
         _opts.shouldNot.forEach((p: string) => {
           const property = getProperties(bodyParameter)?.[p];
           if (property) {
-            var isPropertyBoolean = false;
-            var isPropertyImmutable = false;
+            let isPropertyBoolean = false;
+            let isPropertyImmutable = false;
 
             if (property["readOnly"] && property["readOnly"] === true)
             {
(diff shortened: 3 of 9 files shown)


common/changes/@microsoft.azure/openapi-validator-rulesets/rmanda-fix_rpc_patch_v1_02_2023-05-12-05-42.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
```
konrad-jamrozik, line 2: maybe prefix this file with `rkmanda` not `rmanda` to avoid confusion on user names.

docs/un-supported-patch-properties.md
```diff
@@ -30,4 +30,4 @@ July 07, 2022
 
 ## How to fix the violation
 
-Considering removing the name, location, or type in the patch request body.
+Consider either removing the top level properties - id, name and type, from the patch request body or mark them as readOnly. For the top level location property for tracked resources, consider either removing it from the request body of the Patch operation or mark it as immutable using the x-ms-mutability property and values as "create" and "read".
```
konrad-jamrozik, line 33:
```suggestion
Consider either removing the top level properties - id, name and type, from the patch request body, or mark them as readOnly. For the top level location property for tracked resources, consider either removing it from the request body of the Patch operation or mark it as immutable using the x-ms-mutability property and values as "create" and "read".
```

docs/un-supported-patch-properties.md
```diff
@@ -14,7 +14,7 @@ ARM OpenAPI(swagger) specs
 
 ## Output Message
 
-The patch operation body parameter schema should not contains property {propertyName}.
+The patch operation body parameter schema should not contain top level "id", "name", "type", "location" as writable properties. This is because these properties are expected to be either readOnly or immutable.
```
konrad-jamrozik, line 17:
```suggestion
The patch operation body parameter schema should not contain top level "id", "name", "type", or "location" as writable properties. This is because these properties are expected to be either readOnly or immutable.
```

docs/un-supported-patch-properties.md
konrad-jamrozik, line 33: What does `for tracked resources` mean here? Can there be a top-level "location" property for untracked resources?

packages/rulesets/src/spectral/functions/validate-patch-body-param-properties.ts
```diff
@@ -46,19 +46,39 @@ export const validatePatchBodyParamProperties = createRulesetFunction<unknown, O
...
+            let isPropertyBoolean = false;
+            let isPropertyImmutable = false;
+
+            if (property["readOnly"] && property["readOnly"] === true)
+            {
+              isPropertyBoolean = true;
+            }
+
+            if (property["x-ms-mutability"] && Array.isArray(property["x-ms-mutability"])) {
+                const schemeArray = property["x-ms-mutability"];
```
konrad-jamrozik, line 68: `schemeArray` or `schemaArray` ?

packages/rulesets/src/spectral/functions/validate-patch-body-param-properties.ts
```diff
@@ -46,19 +46,39 @@ export const validatePatchBodyParamProperties = createRulesetFunction<unknown, O
...
-              path: [...path, "parameters", index],
-            })
+          const property = getProperties(bodyParameter)?.[p];
+          if (property) {
+            let isPropertyBoolean = false;
+            let isPropertyImmutable = false;
+
+            if (property["readOnly"] && property["readOnly"] === true)
+            {
+              isPropertyBoolean = true;
```
konrad-jamrozik, line 64: Inconsistent indentation: sometimes I see indentation of 4, sometimes of 2.

Similar problem with bracketing style. Sometimes I see opening `{` on the same line, sometimes on the next one.

Let's be consistent. I think this codebase has built-in formatter, let's use it. Here it is:
https://github.com/Azure/azure-openapi-validator/blob/aa7c987f73d163cb5f2810a5323a61bb869bb4e5/.vscode/settings.json#L1C1-L8

packages/rulesets/src/spectral/functions/validate-patch-body-param-properties.ts
```diff
@@ -46,19 +46,39 @@ export const validatePatchBodyParamProperties = createRulesetFunction<unknown, O
...
+                if (!schemeArray.includes("update"))
+                {
+                    isPropertyImmutable = true;
+                }
+            }
+
+            if (!isPropertyBoolean && !isPropertyImmutable)
+            {
+              errors.push({
+                message: `The patch operation body parameter schema should not contain property ${p}.`,
```
konrad-jamrozik, line 78: Can we provide more information here? Like:

`message: 'The patch operation body parameter schema should not contain property ${p}, because the property is not readable nor immutable.'`,

Also, I am confused: the doc for the rule says that the property should not be included in parameters if `it is read-only or it is immutable`, but this code makes me think the property should not be included in parameters `unless it is read-only or it is immutable`. Which one is it?

Maybe instead `if (!isPropertyBoolean && !isPropertyImmutable)`
we should have `if (isPropertyReadOnly || isPropertyImmutable)` ? 

Then it would make sense to me: if a property is read-only or immutable, we notify the user it should not be in the params.

packages/rulesets/src/spectral/functions/validate-patch-body-param-properties.ts
```diff
@@ -46,19 +46,39 @@ export const validatePatchBodyParamProperties = createRulesetFunction<unknown, O
...
+
+            if (property["x-ms-mutability"] && Array.isArray(property["x-ms-mutability"])) {
+                const schemeArray = property["x-ms-mutability"];
+                if (!schemeArray.includes("update"))
+                {
+                    isPropertyImmutable = true;
+                }
+            }
+
+            if (!isPropertyBoolean && !isPropertyImmutable)
```
konrad-jamrozik, line 75:
```suggestion
            if (!(isPropertyReadOnly || isPropertyImmutable))
```

I think this is easier to reason about.

packages/rulesets/src/spectral/functions/validate-patch-body-param-properties.ts
konrad-jamrozik, line 64:
```suggestion
              isPropertyReadOnly = true;
```

konrad-jamrozik (changes requested):

common/changes/@microsoft.azure/openapi-validator-rulesets/rmanda-fix_rpc_patch_v1_02_2023-05-12-05-42.json
rkmanda: the nameis automatically derived based on the branch name used. I will use my github handle next time in the branch name. will leave this as is for now

docs/un-supported-patch-properties.md
rkmanda: hmm including them as top-level would not require them to be writable.

rkmanda: The properties should either not exist in the patch request body or if they are specified, they must be marked as readOnly or immutable

docs/un-supported-patch-properties.md
rkmanda: fixed

rkmanda: "location" property can only exist for tracked resources. Tracked resources are resources that are stored by ARM in its registry. A resource is a representation of an object that follows the REST semantics

rkmanda: I edited the description to make it convey the meaning better

packages/rulesets/src/spectral/functions/validate-patch-body-param-properties.ts
rkmanda: copy pasted var names :)

packages/rulesets/src/spectral/functions/validate-patch-body-param-properties.ts
rkmanda: fixed

packages/rulesets/src/spectral/functions/validate-patch-body-param-properties.ts
rkmanda: I felt the other way looks easier :)

rkmanda: fixed anyway

packages/rulesets/src/spectral/functions/validate-patch-body-param-properties.ts
```diff
@@ -46,19 +46,36 @@ export const validatePatchBodyParamProperties = createRulesetFunction<unknown, O
...
+            if (property["x-ms-mutability"] && Array.isArray(property["x-ms-mutability"])) {
+              const schemaArray = property["x-ms-mutability"];
+              if (!schemaArray.includes("update")) {
+                isPropertyImmutable = true;
+              }
+            }
+
+            if (!(isPropertyReadOnly || isPropertyImmutable)) {
+              errors.push({
+                message: `Mark the top level property - ${p} specified in the patch operation body as readOnly or immutable. These properties are not patchable.`,
```
konrad-jamrozik, line 75:
```suggestion
                message: `Mark the top-level property, ${p}, specified in the patch operation body as readOnly or immutable. These properties are not patchable.`,
```

I think in the origignal form it can render like this, e.g. for property `location`:

> Mark the top level property - location specified in the patch operation body as readOnly or immutable. 

Which might be confusin.

packages/rulesets/src/spectral/functions/validate-patch-body-param-properties.ts
konrad-jamrozik: OK, I think a little bit too many nested negations threw me off ;) In particular the `_opts.shouldNot` is a bit misleading. Because the property **can** be included even if it **should not**. But if this is the case, then it has to be `readOnly || immutable`.

Btw. maybe we could clarify the intent of `if (!(isPropertyReadonly || isPropertyImmutable))` by extracting a method, like `if (isPropertyWritable())` ? One negation less :)

docs/un-supported-patch-properties.md
konrad-jamrozik, line 17:
```suggestion
The patch operation body parameter schema should not contain top-level "id", "name", "type", "location" as writable properties. To fix this, either remove the offending properties, or mark them as readOnly or immutable.
```

I think this would convey what is going a bit more naturally, based on our discussion.

konrad-jamrozik (approved):
Approved with few additional minor comments. Thanks! :)

packages/rulesets/src/spectral/functions/validate-patch-body-param-properties.ts
konrad-jamrozik: I have one more idea here, as explained [in this comment](https://github.com/Azure/azure-openapi-validator/pull/513#discussion_r1196090530).

packages/rulesets/src/spectral/functions/validate-patch-body-param-properties.ts
bdefoy: Could also put it in quotes

bdefoy (approved):

docs/un-supported-patch-properties.md
bdefoy: Could also put it in quotes

rkmanda: I felt the "How to ix the violation was where we tell what to do to fix the violation so left the Output message to only reflect an explanation of the problem

rkmanda: But I will change it anyway.

rkmanda: Which part in quotes Brett?

bdefoy: I meant to put that reply on https://github.com/Azure/azure-openapi-validator/pull/513#discussion_r1196076146

PR #523, opened 2023-05-18
Add rush prep command

bdefoy:
Add command that runs all the things needed before merging a PR ([reference](https://github.com/Azure/azure-openapi-validator/blob/main/CONTRIBUTING.md#how-to-prepare-for-a-pr-submission-after-you-made-changes-locally)).

### DIFF
--- a/common/config/rush/command-line.json
+++ b/common/config/rush/command-line.json
@@ -57,6 +57,13 @@
       "name": "lint",
       "summary": "Lint projects. Runs `npm run lint` on all projects.",
       "enableParallelism": true
+    },
+    {
+      "commandKind": "global",
+      "name": "prep",
+      "summary": "Run the various commands needed to generate files before merging a pull request.",
+      "shellCommand": "rush update && rush build && rush lint && rush test && rush regen-ruleindex && rush change",
+      "safeForSimultaneousRushProcesses": true
     }
   ]
 }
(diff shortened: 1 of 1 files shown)


rkmanda (approved):
:shipit:

common/config/rush/command-line.json
```diff
@@ -57,6 +57,13 @@
       "name": "lint",
       "summary": "Lint projects. Runs `npm run lint` on all projects.",
       "enableParallelism": true
+    },
+    {
+      "commandKind": "global",
+      "name": "prep",
+      "summary": "Run the various commands needed to generate files before merging a pull request.",
+      "shellCommand": "rush lint && rush update && rush build && rush test && rush regen-ruleindex && rush change",
```
konrad-jamrozik, line 65: Not sure what will happen if we run `rush lint` on a codebase that cannot be successfully built. Probably some unhelpful/misleading error? Maybe we should run `rush lint` after `rush build` ?

Also, will `rush regen-ruleindex` and `rush change` properly do nothing, as appropriate?

In any case, this is not a blocker. We can tweak this command later on.

bdefoy: I just tried it and `rush lint` does a build before running the linter, so it caught there was a build issue. However, the output from `rush build` showed me exactly what the build error was, while the `rush lint` output did not. So, I agree the build should be run first. Reverting my last commit.

bdefoy: > Also, will rush regen-ruleindex and rush change properly do nothing, as appropriate?

Yes, `rush regen-ruleindex` will always generate the rule index from all the .md files, but if there are no changes in the .md files, there will be no changes in the rule index. Also, if there already exists a rush changelog file, `rush change` will tell you that and give you an option (the default option) to skip generating another one.

PR #527, opened 2023-05-25
removed RPC-Async-V1-06 and extra check which is not required as part of CreateOperationAsyncResponseValidation

tejaswiMinnu:
Removed - RPC-Async-V1-06 rule and its corresponding files and information.
 CreateOperationAsyncResponseValidation - has a false positive which has a check-> 'final-state-via' property set to 'azure-async-operation' 
Removed the check which is not required as the part of "CreateOperationAsyncResponseValidation" with this PR.


docs/rules.md
```diff
@@ -925,6 +925,12 @@ This rule is to check if the schemas used by private endpoint conform to the com
 
 Please refer to [private-endpoint-resource-schema-validation.md](./private-endpoint-resource-schema-validation.md) for details.
 
+### PropertiesTypeObjectNoDefinition
```
visingla-ms, line 928: Sometimes also called "Free-form Objects", if we want to add this bit in description.

visingla-ms (approved):

packages/rulesets/src/native/legacyRules/CreateOperationAsyncResponseValidation.ts
```diff
@@ -45,17 +45,6 @@ rules.push({
           location: path,
         }
```
rkmanda, line 46: remove this as well

rkmanda (changes requested):
:clock1:

bdefoy (approved):

rkmanda (approved):
:shipit:

PR #531, opened 2023-06-01
Add support for `is-staging-run` which disables staging-only Spectral rules in production runs. Update relevant doc on rule disabling.

konrad-jamrozik:
Add support for `is-staging-run` AutoRest argument which disables Spectral rules if given rule has property `stagingOnly: true` and `is-staging-run` is not passed or not set to true.

This feature makes it possible to deploy LintDiff changes to production without enabling rules that are not yet ready to be deployed to production, while still keeping them enabled in staging runs to verify their correctness. 

This PR also:
- updates relevant doc on rule disabling,
- simplifies the `utils.ts / buildLinter` method,
- does few minor source code symbols rename refactorings.

This PR requires a companion PR in `openapi-alps` repo to actually take advantage of the introduced flag. Here it is:
- https://devdiv.visualstudio.com/DefaultCollection/DevDiv/_git/openapi-alps/pullrequest/475702

Following PR immediately takes advantage of changes introduced in this PR:
- https://github.com/Azure/azure-openapi-validator/pull/532

## Testing done

I ran LintDiff locally per the docs, using following cmd lines, differing by usage of `--is-staging-run=`: absent, `false` and `true`. I temporarily added `stagingOnly: true` to few rule definitions. The behavior was always as expected, disabling rules with `stagingOnly: true` when `--is-staging-run` was not passed or not set to `true`.

``` bash
autorest --v3 --spectral --azure-validator --use=./packages/azure-openapi-validator/autorest --tag=package-2022-09-preview ../specs-pr/specification/containerservice/resource-manager/Microsoft.ContainerService/fleet/readme.md

autorest --v3 --spectral --azure-validator --use=./packages/azure-openapi-validator/autorest --is-staging-run=false --tag=package-2022-09-preview ../specs-pr/specification/containerservice/resource-manager/Microsoft.ContainerService/fleet/readme.md

autorest --v3 --spectral --azure-validator --use=./packages/azure-openapi-validator/autorest --is-staging-run=true --tag=package-2022-09-preview ../specs-pr/specification/containerservice/resource-manager/Microsoft.ContainerService/fleet/readme.md
```

### DIFF
--- a/packages/azure-openapi-validator/autorest/src/spectral-plugin-func.ts
+++ b/packages/azure-openapi-validator/autorest/src/spectral-plugin-func.ts
@@ -16,6 +16,7 @@ import { cachedFiles } from "."
 export async function spectralPluginFunc(initiator: IAutoRestPluginInitiator): Promise<void> {
   const files: string[] = (await initiator.ListInputs()).filter((f) => !isCommonTypes(f))
   const openapiType: string = await getOpenapiTypeStr(initiator)
+  const isStagingRun: boolean = await initiator.GetValue("is-staging-run")
 
   const readFile = async (fileUri: string) => {
     let file = cachedFiles.get(fileUri)
@@ -30,7 +31,9 @@ export async function spectralPluginFunc(initiator: IAutoRestPluginInitiator): P
   }
 
   const resolvedOpenapiType: OpenApiTypes = getOpenapiType(openapiType)
-  const ruleset: Ruleset = await getRuleSet(resolvedOpenapiType)
+  const [ruleset, namesOfRulesInStagingOnly]: [Ruleset, string[]] = await getRuleSet(resolvedOpenapiType)
+
+  ifNotStagingRunThenDisableStagingOnlyRules(isStagingRun, initiator, ruleset, namesOfRulesInStagingOnly)
 
   printRuleNames(initiator, ruleset, resolvedOpenapiType)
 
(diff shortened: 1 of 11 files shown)


konrad-jamrozik:
Per @rkmanda request, I will adjust this PR to keep Error rules as Errors in staging. I'll investigate a way to instead prevent the Staging LintDiff CI check failing in case any of the rules fail.

packages/azure-openapi-validator/autorest/src/spectral-plugin-func.ts
```diff
@@ -80,6 +83,30 @@ export async function spectralPluginFunc(initiator: IAutoRestPluginInitiator): P
...
+    Object.values(ruleset.rules).forEach((rule) => {
+      // Note that DiagnosticsSeverity doesn't have a value for "off", and so we cannot simply check against it.
+      // The logic is more nuanced, where "off" ends up having value of -1 in DiagnosticSeverity.
+      // This is because function "getRuleSet" defined in this file calls into "set severity" in rule.ts which
+      // has the special handling logic. Source at:
+      // https://github.com/stoplightio/spectral/blob/44c40e2b7c9ea6222054da8700049b0307cc5f8b/packages/core/src/ruleset/rule.ts#L121
+      // where "off" is defined as -1 per:
+      // https://github.com/stoplightio/spectral/blob/44c40e2b7c9ea6222054da8700049b0307cc5f8b/packages/ruleset-migrator/src/transformers/rules.ts#L39
+      if ([DiagnosticSeverity.Warning, DiagnosticSeverity.Error].every((severity) => rule.severity != severity))
+        rule.severity = DiagnosticSeverity.Error
```
weshaggard, line 105: Don't we have some rules that are never supposed to be Error? In such a case won't we have a lot of extra noise here because of this? 

Do we have any options that can mark a new rule as staging or some other indication that doesn't blindly promote everything to error?

konrad-jamrozik: Re

> Don't we have some rules that are never supposed to be Error? In such a case won't we have a lot of extra noise here because of this?

From what I see [all rules are either warning or error](https://github.com/search?q=repo%3AAzure%2Fazure-openapi-validator+%22severity%3A+%22&type=code&p=1). There is one hit against severity of [recommended](https://github.com/Azure/azure-openapi-validator/blob/ce51affdc66115457bba7e7d5860ead92a7f92b4/packages/azure-openapi-validator/autorest/src/spectral-plugin-func.ts#LL178C59-L178C59) but this appears to be some ruleset context severity, which I think is most likely unused.

To get rid of the extra noise I could enable the rules as `Warning`s instead of `Error`s. Or  don't stress over it too much - it is staging after all, and we are planning to prevent it from ever becoming red.

Re

> Do we have any options that can mark a new rule as staging or some other indication that doesn't blindly promote everything to error?

I don't think you can do this, because of the spectral source code, as mentioned in this comment:

``` typescript
      // Note that DiagnosticsSeverity doesn't have a value for "off", and so we cannot simply check against it.
      // The logic is more nuanced, where "off" ends up having value of -1 in DiagnosticSeverity.
      // This is because function "getRuleSet" defined in this file calls into "set severity" in rule.ts which
      // has the special handling logic. Source at:
      // https://github.com/stoplightio/spectral/blob/44c40e2b7c9ea6222054da8700049b0307cc5f8b/packages/core/src/ruleset/rule.ts#L121
      // where "off" is defined as -1 per:
      // https://github.com/stoplightio/spectral/blob/44c40e2b7c9ea6222054da8700049b0307cc5f8b/packages/ruleset-migrator/src/transformers/rules.ts#L39
```

konrad-jamrozik: I ended up applying this approach by stripping out `stagingOnly` just before the object is passed as an argument to `Ruleset` ctor. Later on, in the ctored `Ruleset`, the rules that were `stagingOnly` have their severity set to `off`.

weshaggard (approved):

PR #518, opened 2023-05-16
Add new linter rule promotion strategy

bdefoy:
Updating the contribution document to define a strategy for promoting new linter rules to the production pipeline.

PR #532, opened 2023-06-01
Set recently introduced ARM Spectral rules to `stagingOnly`, so they run only in Staging, not Prod.

konrad-jamrozik:
This PR is setting to `stagingOnly` all rules added since May 10th as this was the most recent deployment of ruleset:
- https://www.npmjs.com/package/@microsoft.azure/openapi-validator-rulesets?activeTab=versions

To figure out which rules needed to be set to `stagingOnly`, I looked at all changes made to `az-arm.ts` after May 10th, per:
- https://github.com/Azure/azure-openapi-validator/commits/main/packages/rulesets/src/spectral/az-arm.ts

This PR **cannot be merged** until the following PRs are deployed to production:

- https://github.com/Azure/azure-openapi-validator/pull/531
- https://devdiv.visualstudio.com/DefaultCollection/DevDiv/_git/openapi-alps/pullrequest/475702

Update: Prerequisites deployed via:
- https://dev.azure.com/azure-sdk/internal/_releaseProgress?_a=release-environment-logs&releaseId=23317&environmentId=28948
- https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=7892511&view=results
- https://github.com/Azure/azure-rest-api-specs-pipeline/commit/1694613959f06dc1e3f448ca77f4e408512a7433

### DIFF
--- a/packages/rulesets/generated/spectral/az-arm.js
+++ b/packages/rulesets/generated/spectral/az-arm.js
@@ -2581,6 +2581,7 @@ const ruleset = {
         DeleteResponseCodes: {
             description: "Synchronous DELETE must have 200 & 204 return codes and LRO DELETE must have 202 & 204 return codes.",
             severity: "error",
+            stagingOnly: true,
             message: "{{error}}",
             resolved: true,
             formats: [oas2],
@@ -2614,6 +2615,7 @@ const ruleset = {
         AvoidAdditionalProperties: {
             description: "The use of additionalProperties is not allowed except for user defined tags on tracked resources.",
             severity: "error",
+            stagingOnly: true,
             message: "{{description}}",
             resolved: true,
             formats: [oas2],
@@ -2625,6 +2627,7 @@ const ruleset = {
         PropertiesTypeObjectNoDefinition: {
             description: "Properties with type:object that don't reference a model definition are not allowed. ARM doesn't allow generic type definitions as this leads to bad customer experience.",
             severity: "error",
+            stagingOnly: true,
             message: "{{error}}",
             resolved: true,
             formats: [oas2],
@@ -3030,6 +3033,7 @@ const ruleset = {
             description: "Service-defined (reserved) resource names must be represented as an enum type with modelAsString set to true, not as a static string in the path.",
             message: "{{error}}",
             severity: "error",
+            stagingOnly: true,
             resolved: true,
             formats: [oas2],
             given: ["$[paths,'x-ms-paths']"],
@@ -3055,6 +3059,7 @@ const ruleset = {
             description: "The get operations endpoint must only be at the tenant level.",
             message: "{{error}}",
             severity: "error",
+            stagingOnly: true,
             resolved: true,
             formats: [oas2],
             given: "$.[paths,'x-ms-paths']",
--- a/packages/rulesets/generated/spectral/az-arm.js
+++ b/packages/rulesets/generated/spectral/az-arm.js
@@ -3071,6 +3071,7 @@ const ruleset = {
             description: "This is a rule introduced to validate if provisioningState property is set to readOnly or not.",
             message: "{{error}}",
             severity: "off",
+            stagingOnly: true,
             resolved: true,
             formats: [oas2],
             given: ["$[paths,'x-ms-paths'].*.*.responses.*.schema"],
--- a/packages/rulesets/src/spectral/az-arm.ts
+++ b/packages/rulesets/src/spectral/az-arm.ts
@@ -160,6 +160,7 @@ const ruleset: any = {
     DeleteResponseCodes: {
       description: "Synchronous DELETE must have 200 & 204 return codes and LRO DELETE must have 202 & 204 return codes.",
       severity: "error",
+      stagingOnly: true,
       message: "{{error}}",
       resolved: true,
       formats: [oas2],
@@ -203,6 +204,7 @@ const ruleset: any = {
     AvoidAdditionalProperties: {
       description: "The use of additionalProperties is not allowed except for user defined tags on tracked resources.",
       severity: "error",
+      stagingOnly: true,
       message: "{{description}}",
       resolved: true,
       formats: [oas2],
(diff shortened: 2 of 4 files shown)


packages/rulesets/generated/spectral/az-arm.js
```diff
@@ -2672,7 +2672,7 @@ const ruleset = {
         },
         DeleteResponseCodes: {
             description: "Synchronous DELETE must have 200 & 204 return codes and LRO DELETE must have 202 & 204 return codes.",
-            severity: "error",
+            severity: "off",
```
weshaggard, line 2675: Is severity strict? In other words could set set this to a value like staging? See my comment in your other PR https://github.com/Azure/azure-openapi-validator/pull/531#discussion_r1213756664.

If we cannot do that is there some way to add a tag or keyword of some sort extra metadata in this configuration to identify this as staging?

konrad-jamrozik: I think we cannot set it to `staging`, per [my comment here](https://github.com/Azure/azure-openapi-validator/pull/531/files/6ab2d5a466c4844216f17e9acdfc3ff2bdd41d01#r1213765233).

I identified it as staging by adding a comment in the source of truth typescript file: [az-arm.ts](https://github.com/Azure/azure-openapi-validator/pull/532/files#diff-4c1382203db84bcd9df61a5bbf90823d0e1f39a833e8eaa1a5be96ca4a4e9b61). The file you are commenting on, `az-arm.js`, is generated from that `.ts` file upon `rush build`.

weshaggard: Yes but can that comment be turned into some other tag field we can filter on?

Another possibility is to split out the staging rules into their own rule set list. Instead of being in az-arm.ts they could be az-staging.ts and then we can include or not include that set based on the -is-staging flag at the higher level at, https://github.com/Azure/azure-openapi-validator/blob/main/packages/rulesets/src/index.ts#L10. That way the promotion procedure is to move from az-staging.js to az-staging.ts.

konrad-jamrozik: > Yes but can that comment be turned into some other tag field we can filter on?

I am unsure about that, because from what I see there is a type conversion going on, from the ruleset yaml into a type provided by the spectral library, [here](https://github.com/Azure/azure-openapi-validator/blob/6ab2d5a466c4844216f17e9acdfc3ff2bdd41d01/packages/azure-openapi-validator/autorest/src/spectral-plugin-func.ts#L205):
`return new Ruleset(ruleset, { severity: "recommended" })`

> Another possibility is to split out the staging rules into their own rule set list. 

I see some complications here. For example, we would need separate staging rules for arm and data plane, and then promoting the rules from staging would mean moving it them over to a different file, resulting in large diff.

How about we stick to the comment for now and iteratively improve the process later on? Note we have a draft PR where we are working on coming up with the rule promotion strategy, here:
- https://github.com/Azure/azure-openapi-validator/pull/518

weshaggard: There has to be a more structured approach than relying on a comment and then force turning on all rules as errors when we want to run staging. I'm sure we can figure out something that will allow us to programmatically determine the set of rules that are in staging.

packages/rulesets/generated/spectral/az-arm.js
```diff
@@ -1669,33 +1669,6 @@ const provisioningStateSpecifiedForLROPatch = (patchOp, _opts, ctx) => {
     return errors;
 };
 
-const LROPostFinalStateViaProperty = (postOp, _opts, ctx) => {
```
konrad-jamrozik, line 1672: FYI I am not yet sure why these rules got deleted, probably some missed `rush build` command by someone else. I'll investigate and fix.

konrad-jamrozik: OK this is definitely missing `rush build`. I'll have a PR for that in a moment.

konrad-jamrozik: Fixed here:
- https://github.com/Azure/azure-openapi-validator/pull/540

packages/rulesets/generated/spectral/az-arm.js
konrad-jamrozik: Changes applied. Ready for review. Details in:
- #531

packages/rulesets/generated/spectral/az-arm.js
konrad-jamrozik: The problem is gone after rebasing on `main` and doing `rush build`.

packages/rulesets/src/spectral/az-arm.ts
```diff
@@ -758,7 +763,7 @@ const ruleset: any = {
     ProvisioningStateMustBeReadOnly: {
       description: "This is a rule introduced to validate if provisioningState property is set to readOnly or not.",
       message: "{{error}}",
-      severity: "off", // See https://github.com/Azure/azure-sdk-tools/issues/6071#issuecomment-1535560188
+      severity: "off", // See https://github.com/Azure/azure-sdk-tools/issues/6191#issuecomment-1571334585
```
weshaggard, line 766: I realize this is a separate issue but should we use the strategy of demoting this to a staging only rule so we can still run it if needed for testing?

weshaggard (approved):

packages/rulesets/src/spectral/az-arm.ts
konrad-jamrozik: I thought so too, but my read from the linked comment is there is no need to run this in staging: we know this is rule is borked and how to fix it. I guess once it is fixed, then we can re-enable it for staging.

konrad-jamrozik: Done.

konrad-jamrozik:
Rolled out to staging with:
- https://dev.azure.com/azure-sdk/internal/_releaseProgress?_a=release-environment-logs&releaseId=23321&environmentId=28952
and to prod with:
- https://dev.azure.com/azure-sdk/internal/_releaseProgress?_a=release-environment-logs&releaseId=23322&environmentId=28953

PR #535, opened 2023-06-02
merging the relevant checks from DeleteOperationAsyncResponseValidation into delete-response-codes and removing DeleteOperationAsyncResponseValidation

rkmanda:
The DeleteOperationAsyncResponseValidation allows 200 to be a valid status code which conflicts with the rule in delete-response-codes which is more restrictive and does not allow 200. Based on an internal email thread with the API reviewers we concluded that for async delete operations we do not want to let authors specify a 200 status code. This PR applies this correction. 
Also there were some relevant checks missing from delete-response-codes which existed in DeleteOperationAsyncResponseValidation like the criteria to consider an API as async and flagging an error when the x-ms-long-running-operation extension is not applied. These validations have now been incorporated from DeleteOperationAsyncResponseValidation  into delete-response-codes

Addressing task https://msazure.visualstudio.com/One/_workitems/edit/23293646

and https://msazure.visualstudio.com/One/_workitems/edit/23293632


packages/rulesets/generated/spectral/az-arm.js
```diff
@@ -1448,20 +1448,42 @@ const consistentPatchProperties = (patchOp, _opts, ctx) => {
 
 const SYNC_DELETE_RESPONSES = ["200", "204", "default"];
 const LR_DELETE_RESPONSES = ["202", "204", "default"];
-const SYNC_ERROR = "Synchronous delete operations must have responses with 200, 204, and default return codes. They also must have no other response codes.";
-const LR_ERROR = "Long-running (LRO) delete operations must have responses with 202, 204, and default return codes. They also must have no other response codes.";
+const SYNC_ERROR = "Synchronous delete operations must have responses with 200, 204 and default return codes. They also must have no other response codes.";
```
ravimeda, line 1451: Curious - What are the "default return codes"?

packages/rulesets/generated/spectral/az-arm.js
```diff
@@ -1448,20 +1448,42 @@ const consistentPatchProperties = (patchOp, _opts, ctx) => {
...
+const LR_ERROR = "Long-running (LRO) delete operations must have responses with 202, 204 and default return codes. They also must have no other response codes.";
```
ravimeda, line 1452: Nit - In the "Long-running (LRO)", I think the abbreviation is unnecessary.

packages/rulesets/generated/spectral/az-arm.js
bdefoy: This is referring to the default response. See the Default Response section on this page: https://swagger.io/docs/specification/describing-responses/

packages/rulesets/generated/spectral/az-arm.js
```diff
@@ -1448,20 +1448,42 @@ const consistentPatchProperties = (patchOp, _opts, ctx) => {
...
+const EmptyResponse_ERROR = "Delete operation response codes MUST be non-empty. It MUST have response codes 200, 204 and default if it is sync OR 202, 204 and default if it is long running.";
```
ravimeda, line 1453: Nit - About "MUST" and "OR", I see you uppercased to help the user notice the issue. I would suggest lowercase.

For example - "The response codes for a delete operation must not be empty. The response code should be either 200 or 204, if the operation is synchronous. If the operation is async and long running, then the response code should be either 202 or 204."

packages/rulesets/src/spectral/functions/delete-response-codes.ts
```diff
@@ -15,13 +17,38 @@ export const DeleteResponseCodes = (deleteOp: any, _opts: any, ctx: any) => {
...
-    return []
+  if (!deleteOp?.responses || deleteOp?.responses.length == 0) {
+    errors.push({
+      message: EmptyResponse_ERROR,
+      path: path,
+    })
+    return errors
   }
 
   const responses = Object.keys(deleteOp?.responses ?? {})
```
bdefoy, line 28: I think it should be this now that I'm seeing this line again

```suggestion
  const responses = Object.keys(deleteOp?.responses ?? [])
```

packages/rulesets/src/spectral/functions/delete-response-codes.ts
```diff
@@ -15,13 +17,38 @@ export const DeleteResponseCodes = (deleteOp: any, _opts: any, ctx: any) => {
...
 
-  if (!deleteOp?.responses) {
-    return []
+  if (!deleteOp?.responses || deleteOp?.responses.length == 0) {
+    errors.push({
+      message: EmptyResponse_ERROR,
+      path: path,
+    })
+    return errors
   }
```
bdefoy, line 26: I think this can be removed since I think lines 28 through 36 cover these cases
```suggestion
```

rkmanda: I forgot to remove the OR condition! But I see that both conditions are covered below, now! Tx!

packages/rulesets/generated/spectral/az-arm.js
rkmanda: removed

packages/rulesets/generated/spectral/az-arm.js
rkmanda: modified casing

packages/rulesets/src/spectral/functions/delete-response-codes.ts
rkmanda: removed redundant check

ravimeda:
Looks good to me.

ravimeda (approved):

tejaswiMinnu (approved):

PR #534, opened 2023-06-02
Allow "format: uri" in dataplane

mikekistler:
This PR makes a small change to the SchemaTypeAndFormat rule to allow `format: uri` for data plane REST APIs. This rule only runs for data plane, and the corresponding rule for ARM, [ValidFormats](https://github.com/Azure/azure-openapi-validator/blob/main/docs/valid-formats.md) already allows `format: uri`.  Autorest accepts `format: uri` as equivalent to `format: url`.

Tests and documentation have been updated.

PR #540, opened 2023-06-07
Run leftover `rush build` and apply fixes so `rush regen-ruleindex` works as intended

konrad-jamrozik:
This PR applies fixes that should have been done as part of these PRs:
- #534
- #537


mikekistler (approved):
Looks good! 👍

PR #541, opened 2023-06-07
GetCollectionOnlyHasValueAndNextLink 

tejaswiMinnu:
This rule dint check if the path is default or not which resulted false positve.
Added a condition which checks if path is default.
WI - https://msazure.visualstudio.com/One/_workitems/edit/22895869

### DIFF
--- a/packages/rulesets/generated/spectral/az-arm.js
+++ b/packages/rulesets/generated/spectral/az-arm.js
@@ -2661,7 +2661,7 @@ const ruleset = {
             severity: "error",
             resolved: true,
             formats: [oas2],
-            given: "$[paths,'x-ms-paths'][?(!@property.endsWith('}') && !@property.endsWith('operations'))][get].responses.200.schema.properties",
+            given: "$[paths,'x-ms-paths'][?(!@property.endsWith('}') && !@property.endsWith('operations') && !@property.endsWith('default'))][get].responses.200.schema.properties",
             then: {
                 function: getCollectionOnlyHasValueAndNextLink,
             },
--- a/packages/rulesets/src/spectral/az-arm.ts
+++ b/packages/rulesets/src/spectral/az-arm.ts
@@ -265,7 +265,8 @@ const ruleset: any = {
       severity: "error",
       resolved: true,
       formats: [oas2],
-      given: "$[paths,'x-ms-paths'][?(!@property.endsWith('}') && !@property.endsWith('operations'))][get].responses.200.schema.properties",
+      given:
+        "$[paths,'x-ms-paths'][?(!@property.endsWith('}') && !@property.endsWith('operations') && !@property.endsWith('default'))][get].responses.200.schema.properties",
       then: {
         function: getCollectionOnlyHasValueAndNextLink,
       },
--- a/packages/rulesets/src/spectral/test/get-collection-only-has-value-nextlink.test.ts
+++ b/packages/rulesets/src/spectral/test/get-collection-only-has-value-nextlink.test.ts
@@ -12,21 +12,54 @@ test("GetCollectionOnlyHasValueAndNextLink should find no errors when get collec
   const myOpenApiDocument = {
     swagger: "2.0",
     paths: {
-      "/api/Paths": {
-        get: {
-          operationId: "Noun_Get",
-          responses: {
-            "200": {
-              schema: {
-                $ref: "#/definitions/ListResult",
+      "/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.Network/dnsZones/{zoneName}/dnssecConfigs/default":
+        {
+          get: {
+            operationId: "Noun_Get",
+            responses: {
+              "200": {
+                schema: {
+                  $ref: "#/definitions/ListResult",
+                },
               },
-            },
-            default: {
-              description: "Unexpected error",
+              default: {
+                description: "Unexpected error",
+              },
             },
           },
         },
-      },
+      "/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.Network/dnsZones/{zoneName}/dnssecConfigs/operations":
+        {
+          get: {
+            operationId: "Noun_Get",
+            responses: {
+              "200": {
+                schema: {
+                  $ref: "#/definitions/ListResult",
+                },
+              },
+              default: {
+                description: "Unexpected error",
+              },
+            },
+          },
+        },
+      "/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.Network/dnsZones/{zoneName}/dnssecConfigs/{dnssecConfigsName}":
+        {
+          get: {
+            operationId: "Noun_Get",
+            responses: {
+              "200": {
+                schema: {
+                  $ref: "#/definitions/ListResult",
+                },
+              },
+              default: {
+                description: "Unexpected error",
+              },
+            },
+          },
+        },
     },
     definitions: {
       ListResult: {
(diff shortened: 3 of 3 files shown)


packages/rulesets/src/spectral/functions/get-collection-only-has-value-nextlink.ts
```diff
@@ -1,9 +1,6 @@
 // Verifies that collection get schemas have only the `value` and `nextLink` properties in their models.
 
 export const getCollectionOnlyHasValueAndNextLink = (properties: any, _opts: any, ctx: any) => {
-  if (!properties || typeof properties !== "object") {
```
bdefoy, line 4: Why did this need to be removed?

tejaswiMinnu: These lines never are never executed, we are sending properties object to this function.

bdefoy: I think most of our functions have these checks at the beginning so that in case somehow a bad parameter gets passed in we won't end up with a null reference or some type error. I think we should keep it in here

tejaswiMinnu: But when I leave the condition, as that is dead code unable to cover that in tests and build is failing due to test coverage percentage.

bdefoy (approved):

packages/rulesets/src/spectral/az-arm.ts
```diff
@@ -265,7 +265,8 @@ const ruleset: any = {
       severity: "error",
       resolved: true,
       formats: [oas2],
-      given: "$[paths,'x-ms-paths'][?(!@property.endsWith('}') && !@property.endsWith('operations'))][get].responses.200.schema.properties",
+      given:
+        "$[paths,'x-ms-paths'][?(!@property.endsWith('}') && !@property.endsWith('operations') && !@property.endsWith('default'))][get].responses.200.schema.properties",
```
rkmanda, line 269: >!@property.endsWith('default') [](http://example.com/codeflow?start=98&length=30)

we discussed that this rule needs to work for paths ending with other static strings as well not just default. We also discussed that the way to apply the rule could be by counting the number of segments after the providerNamespace section of the api path. The rule should be applied when the GET is against odd number of segments after the namespace segments and not apply for GETs against even number of segments.

rkmanda (changes requested):
:clock1:

packages/rulesets/src/spectral/az-arm.ts
tejaswiMinnu: Addressed all the static strings.

packages/rulesets/generated/spectral/az-arm.js
```diff
@@ -1511,17 +1511,27 @@ const DeleteResponseCodes = (deleteOp, _opts, ctx) => {
...
-        ];
+const getCollectionOnlyHasValueAndNextLink = (prop, _opts, ctx) => {
+    for (let path of ctx.path) {
+        if (path.includes(".")) {
+            let splitNamespace = path.split(".");
+            if (path.includes("/")) {
+                let segments = splitNamespace[1].split("/");
+                if (segments.length % 2 !== 0) {
+                    return [];
+                }
```
rkmanda, line 1522: add a comment at the top of this code block to indicate why this logic is being added.

packages/rulesets/src/spectral/functions/get-collection-only-has-value-nextlink.ts
```diff
@@ -1,17 +1,27 @@
...
-      {
-        message: "Get endpoints for collections of resources must only have the `value` and `nextLink` properties in their model.",
-      },
-    ]
+export const getCollectionOnlyHasValueAndNextLink = (prop: any, _opts: any, ctx: any) => {
+  for (let path of ctx.path) {
+    if (path.includes(".")) {
+      // Splitting the path by Provider namespace
+      let splitNamespace = path.split(".")
+      // finiding number of segments after namespace
```
rkmanda, line 8: >finiding [](http://example.com/codeflow?start=9&length=8)

finding

packages/rulesets/src/spectral/functions/get-collection-only-has-value-nextlink.ts
```diff
@@ -1,17 +1,27 @@
...
+      if (path.includes("/")) {
+        let segments = splitNamespace[1].split("/")
+        if (segments.length % 2 !== 0) {
```
rkmanda, line 11: >  if (segments.length % 2 !== 0) { [](http://example.com/codeflow?start=6&length=34)

should this not be segments.length % 2 === 0 ?

packages/rulesets/src/spectral/functions/get-collection-only-has-value-nextlink.ts
rkmanda: lets add this back now and check if there is enough coverage?

packages/rulesets/src/spectral/functions/get-collection-only-has-value-nextlink.ts
```diff
@@ -1,17 +1,27 @@
...
-    return [
-      {
-        message: "Get endpoints for collections of resources must only have the `value` and `nextLink` properties in their model.",
-      },
-    ]
+export const getCollectionOnlyHasValueAndNextLink = (prop: any, _opts: any, ctx: any) => {
+  for (let path of ctx.path) {
+    if (path.includes(".")) {
+      // Splitting the path by Provider namespace
+      let splitNamespace = path.split(".")
```
rkmanda, line 7: > let splitNamespace = path.split(".") [](http://example.com/codeflow?start=5&length=37)

how are you extracting just the part of the path after namespace?

packages/rulesets/src/spectral/functions/get-collection-only-has-value-nextlink.ts
tejaswiMinnu: In the first "if" I divided with dot, so half of the namespaces will be one word which is Network from Microsoft.Network
 After "/" more even values i.e. 4 -> dnsZones/{zoneName}/dnssecConfigs/anystring
Microsoft.Network/dnsZones/{zoneName}/dnssecConfigs/anystring - total count will be odd

packages/rulesets/src/spectral/functions/get-collection-only-has-value-nextlink.ts
tejaswiMinnu: Namespace will have dot right, Microsoft.Namespace so with I am dividing the path.

rkmanda: ok, sorry I didnt read the code properly before. Thinking about this it is possible that some paths have 2 or more namespace names in them, and we should always apply the logic for the last namespace name in the path. CAn u adjust the logic to account for this?

tejaswiMinnu: Added logic to cover if there are multiple provider namespaces.

packages/rulesets/src/spectral/test/get-collection-only-has-value-nextlink.test.ts
```diff
@@ -62,15 +193,21 @@ test("GetCollectionOnlyHasValueAndNextLink should find no errors when get collec
...
   })
 })
 
-test("GetCollectionOnlyHasValueAndNextLink should find errors when get collection schema has properties other than value and nextLink", () => {
+test("GetCollectionOnlyHasValueAndNextLink should find errors when get collection schema has two provider namespaces", () => {
   const myOpenApiDocument = {
     swagger: "2.0",
     paths: {
-      "/api/Paths": {
+      "/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.Network/dnsZones/{zoneName}/dnssecConfigs": {
```
rkmanda, line 210: >Microsoft.Network/dnsZones/{zoneName}/dnssecConfigs": [](http://example.com/codeflow?start=84&length=53)

I am seeing only Microsoft.Network here . which is the other namespace?

tejaswiMinnu: Added

rkmanda (approved):
:shipit:

PR #547, opened 2023-06-14
fix PutRequestResponseSchemeArm for a case where the model has a property without a value 

rkmanda:
PutRequestResponseSchemeArm compares the request and response payloads of the Put operation to check if they match.
A false alarm was reported for this rule by a PR author. The same model was being referenced by the request and the response schemas but the rule was throwing an error indicating that the models are different.

[Add auto rotation param to databases api by viparek · Pull Request #24339 · Azure/azure-rest-api-specs (github.com)](https://github.com/Azure/azure-rest-api-specs/pull/24339/checks?check_run_id=14090301774)

[[DO NOT MERGE] Adding JobAgents API with Identity support to version 2023-05-01-preview by rayboyd30 · Pull Request #24293 · Azure/azure-rest-api-specs (github.com)](https://github.com/Azure/azure-rest-api-specs/pull/24293/checks?check_run_id=14137800398)

ref task : https://msazure.visualstudio.com/One/_workitems/edit/24264914

Issue: The issue is that in the schema comparison logic we are erroring out if the number of properties inside the value of a containing property is 0.
When the length is 0 there isnt anything to compare further and the recursion can be terminated - which is what this fix does.

Test
The test that was added was failing before the fix and it now passes

### DIFF
--- a/packages/rulesets/src/spectral/test/put-request-response-scheme-arm.test.ts
+++ b/packages/rulesets/src/spectral/test/put-request-response-scheme-arm.test.ts
@@ -350,59 +350,6 @@ test("PutRequestResponseScheme for json property with empty value in swagger sho
       },
     },
     definitions: {
-      systemData: {
-        description: "Metadata pertaining to creation and last modification of the resource.",
-        type: "object",
-        readOnly: true,
-        properties: {
-          createdBy: {
-            type: "string",
-            description: "The identity that created the resource.",
-          },
-          createdByType: {
-            type: "string",
-            description: "The type of identity that created the resource.",
-            enum: ["User", "Application", "ManagedIdentity", "Key"],
-            "x-ms-enum": {
-              name: "createdByType",
-              modelAsString: true,
-            },
-          },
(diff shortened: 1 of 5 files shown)


packages/rulesets/src/spectral/test/put-request-response-scheme-arm.test.ts
```diff
@@ -0,0 +1,641 @@
...
+        },
+      },
+    },
+  }
+  return linter.run(myOpenApiDocument).then((results) => {
+    expect(results.length).toBe(0)
+  })
+})
+
+test("PutRequestResponseScheme for json property with empty value in swagger should find no errors. JobAgent.x-ms-arm-id-details is empty below", () => {
```
tejaswiMinnu, line 307: Can we remove extra fields which are not required for this test, it's hard to read.

tejaswiMinnu (approved):

bdefoy (approved):

konrad-jamrozik (approved):

PR #548, opened 2023-06-15
fix verifyArmPath to account for third party RP namespaces

rkmanda:
This PR is to fix a false alarm where some rules like PathContainsResourceType are flagging false alarms when the namespace belongs to third party resource providers. The false alarm is because the verifyArmPath function looks for the term "microsoft" in the path and third party RP namespacds do not start with Microsoft.

To fix this the regex is being adjusted. Specifically the term "Microsoft" is being replaced with the regex \w+ which will allow any word chars in that segment. Retaining the '\.' pattern after that ensures that the segment following the "providers" segment is always a namespace.  

[Task 24144065](https://msazure.visualstudio.com/One/_workitems/edit/24144065): [LintDiff][False positive] PathContainsResourceType

### DIFF
--- a/packages/rulesets/generated/spectral/az-arm.js
+++ b/packages/rulesets/generated/spectral/az-arm.js
@@ -1285,15 +1285,15 @@ function verifyResourceGroupScope(path) {
     return matchAnyPatterns(patterns, path);
 }
 function verifyResourceType(path) {
-    const patterns = [/^.*\/providers\/microsoft\.\w+\/\w+.*/gi];
+    const patterns = [/^.*\/providers\/\w+\.\w+\/\w+.*/gi];
     return matchAnyPatterns(patterns, path);
 }
 function verifyNestResourceType(path) {
     const patterns = [
-        /^.*\/providers\/microsoft\.\w+\/\w+\/{\w+}(?:\/\w+\/(?!default)\w+){1,2}$/gi,
-        /^.*\/providers\/microsoft\.\w+(?:\/\w+\/(default|{\w+})){1,2}(?:\/\w+\/(?!default)\w+)+$/gi,
-        /^.*\/providers\/microsoft\.\w+\/\w+\/(?:\/\w+\/(default|{\w+})){0,3}{\w+}(?:\/{\w+})+.*$/gi,
-        /^.*\/providers\/microsoft\.\w+(?:\/\w+\/(default|{\w+})){0,2}(?:\/\w+\/(?!default)\w+)+\/{\w+}.*$/gi,
+        /^.*\/providers\/\w+\.\w+\/\w+\/{\w+}(?:\/\w+\/(?!default)\w+){1,2}$/gi,
+        /^.*\/providers\/\w+\.\w+(?:\/\w+\/(default|{\w+})){1,2}(?:\/\w+\/(?!default)\w+)+$/gi,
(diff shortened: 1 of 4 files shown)


packages/rulesets/generated/spectral/az-arm.js
```diff
@@ -1285,15 +1285,15 @@ function verifyResourceGroupScope(path) {
...
+    const patterns = [/^.*\/providers\/\w+\.\w+\/\w+.*/gi];
     return matchAnyPatterns(patterns, path);
 }
 function verifyNestResourceType(path) {
     const patterns = [
-        /^.*\/providers\/microsoft\.\w+\/\w+\/{\w+}(?:\/\w+\/(?!default)\w+){1,2}$/gi,
-        /^.*\/providers\/microsoft\.\w+(?:\/\w+\/(default|{\w+})){1,2}(?:\/\w+\/(?!default)\w+)+$/gi,
-        /^.*\/providers\/microsoft\.\w+\/\w+\/(?:\/\w+\/(default|{\w+})){0,3}{\w+}(?:\/{\w+})+.*$/gi,
-        /^.*\/providers\/microsoft\.\w+(?:\/\w+\/(default|{\w+})){0,2}(?:\/\w+\/(?!default)\w+)+\/{\w+}.*$/gi,
+        /^.*\/providers\/\w+\.\w+\/\w+\/{\w+}(?:\/\w+\/(?!default)\w+){1,2}$/gi,
```
ravimeda, line 1293: For learning purpose -
- Is there a limit on the namespace word length?
- Can a namespace have multiple `.` e.g., MyTravels.Oceans.Cruises.Boats ?

ravimeda (approved):

packages/rulesets/generated/spectral/az-arm.js
rkmanda: There is no limit defined in the RPC for namespace names. It is expected to have only 1 dot. This regex was authored many years ago and needs some adjustment to not allow the multiple dots like you noted and possibly few other things. I will doublecheck if there are other rules covering these cases.

AkhilaIlla (approved):

AkhilaIlla:
build is failing!
will need to run "rush change" & add the change logs

rkmanda:
done, tx!

---
In reply to: [1593281581](https://github.com/Azure/azure-openapi-validator/pull/548#issuecomment-1593281581) [](http://example.com/codeflow?ancestors=1593281581,1593281581)

packages/rulesets/generated/spectral/az-arm.js
```diff
@@ -1285,15 +1285,15 @@ function verifyResourceGroupScope(path) {
     return matchAnyPatterns(patterns, path);
 }
 function verifyResourceType(path) {
-    const patterns = [/^.*\/providers\/microsoft\.\w+\/\w+.*/gi];
+    const patterns = [/^.*\/providers\/\w+\.\w+\/\w+.*/gi];
     return matchAnyPatterns(patterns, path);
 }
 function verifyNestResourceType(path) {
     const patterns = [
-        /^.*\/providers\/microsoft\.\w+\/\w+\/{\w+}(?:\/\w+\/(?!default)\w+){1,2}$/gi,
```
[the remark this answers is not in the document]
rkmanda: These were autogenerated after running rush build

PR #571, opened 2023-08-07
Tejaswis/delete rule rpc put v1 11

tejaswiMinnu:
Delete the rule RPC_Put_V1_11


docs/rules.md
```diff
@@ -981,11 +974,11 @@ The request & response('200') schema of the PUT operation must be same.
 
 Please refer to [put-request-response-scheme.md](./put-request-response-scheme.md) for details.
 
-### PutResponseSchemaDescription
```
AkhilaIlla, line 984: >PutResponseSchemaDescription [](http://example.com/codeflow?start=4&length=28)

Why are we deleting this ?

docs/rules.md
```diff
@@ -472,13 +472,6 @@ Location header must be supported for all async operations that return 202.
 
 Please refer to [lro-location-header.md](./lro-location-header.md) for details.
 
-### PostResponseCodes
-
```
AkhilaIlla, line 476: Why are we deleting this ?

tejaswiMinnu: These are rush regen changes, I haven't manually deleted anything in this file.

AkhilaIlla: Okay! I see that https://github.com/Azure/azure-openapi-validator/blob/main/docs/rules.md#postresponsecodes-1 is the right description 
and the one thats getting removed is old

AkhilaIlla (approved):

PR #575, opened 2023-08-15
Make ReservedResourceNamesModelAsEnum a warning and correct the path output

bdefoy:
[ADO PBI 24837788](https://msazure.visualstudio.com/One/_workitems/edit/24837788)

- Correct the output path. Currently pointing to, e.g., `paths./path/.get` but should point to `paths./path/`
- Make it a warning instead of an error


rkmanda (approved):
:shipit:

ravimeda (approved):

PR #580, opened 2023-08-16
Fix severity of the PatchSkuProperty and PatchIdentityProperty rules to warning

rkmanda:
There are some services that do not allow the Sku and identity property to be updated once it is set, so this rule should just be a warning.

### DIFF
--- a/docs/patch-sku-property.md
+++ b/docs/patch-sku-property.md
@@ -2,7 +2,7 @@
 
 ## Category
 
-ARM Error
+ARM Warning
 
 ## Applies to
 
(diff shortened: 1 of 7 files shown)


AkhilaIlla (approved):

docs/patch-sku-property.md
```diff
@@ -14,11 +14,11 @@ ARM OpenAPI(swagger) specs
 
 ## Output Message
 
-The patch operation body parameter schema should contain property 'sku'.
+The patch operation body parameter schema should contain property 'sku' if the service supports updation of the property.
 
 ## Description
 
-RP must implement PATCH for the 'SKU' envelope property if it's defined in the resource model.
+RP should consider implementing Patch for the 'SKU' envelope property if it's defined in the resource model. You may ignore this violation if your service does not allow updation of the Sku property once it is set. In such a case the property must be marked with x-ms-mutability [create, read]
```
AkhilaIlla, line 21: >In such a case the property must be marked with x-ms-mutability [create, read] [](http://example.com/codeflow?start=215&length=78)

I was thinking if we'd need to check/validate this, but its kinda hard to know which service has immutable property

bdefoy (approved):

docs/patch-sku-property.md
rkmanda: yup same thoughts went thru my mind when I was writing it up :) Will park it for later.

PR #583, opened 2023-08-22
Add non fully qualified paths extension resource reg to be excluded f…

AkhilaIlla:
TopLevelResourcesListBySubscription rules checks if all the top-level resources have list by subscription defined (i.e, subscription level get call should exist for all the top-level resources)
Toay the logic doesn't account for the unqualified extension resources. Update the ExtensionResourceReg to consider the  unqualified extension resource paths and exclude them from being considered while building the topLevelResources

### DIFF
--- a/common/changes/@microsoft.azure/openapi-validator-rulesets/akhilailla-ListBySubscriptionForExtensionResourceFalsePositive_2023-08-22-22-01.json
+++ b/common/changes/@microsoft.azure/openapi-validator-rulesets/akhilailla-ListBySubscriptionForExtensionResourceFalsePositive_2023-08-22-22-01.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@microsoft.azure/openapi-validator-rulesets",
+      "comment": "update extensionResourceRegex to also include unqualified paths",
+      "type": "patch"
+    }
+  ],
+  "packageName": "@microsoft.azure/openapi-validator-rulesets"
+}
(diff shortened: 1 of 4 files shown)


packages/rulesets/src/native/utilities/arm-helper.ts
```diff
@@ -55,7 +55,9 @@ export class ArmHelper {
 
   private SpecificResourcePathRegEx = new RegExp("/providers/[^/]+(?:/\\w+/default|/\\w+/{[^/]+})+$", "gi")
 
-  private ExtensionResourceReg = new RegExp(".+/providers/.+/providers/.+$", "gi")
+  private ExtensionResourceFullyQualifiedPathReg = new RegExp(".+/providers/.+/providers/.+$", "gi")
+  private ExtensionResourceReg = new RegExp("^/{", "gi")
```
bdefoy, line 59: Is this enough to ensure the endpoint is an extension resource? Would it be of value to make it something like `^/{\w+}/providers/.+$`?

common/changes/@microsoft.azure/openapi-validator-rulesets/akhilailla-ListBySubscriptionForExtensionResourceFalsePositive_2023-08-22-22-01.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@microsoft.azure/openapi-validator-rulesets",
+      "comment": "update extensionResourceRegex to also include unqualified paths",
```
bdefoy, line 5: I think this might be clearer from the user perspective. Is this accurate for the change?
```suggestion
      "comment": "Add support for extension resource paths that are not fully qualified",
```

rkmanda (changes requested):
:clock1:

bdefoy (approved):

rkmanda (approved):
:shipit:

PR #585, opened 2023-09-02
Added new rule to check ASYNC headers and x-ms-long-running-operation property

tejaswiMinnu:
If responses have headers 'Location' or 'Azure-AsyncOperation', the then Async operations must set 'x-ms-long-running-operation' property set to 'true'

### DIFF
--- a/docs/xms-long-running-operation-property.md
+++ b/docs/xms-long-running-operation-property.md
@@ -1,0 +1,67 @@
+# XmsLongRunningOperationProperty
+
+## Category
+
+ARM Error
+
+## Applies to
+
+ARM OpenAPI(swagger) specs
+
+## Related ARM Guideline Code
+
+- RPC-Async-V1-15
+
+## Output Message
+
+If an operation's (PUT/POST/PATCH/DELETE) responses have `Location` or `Azure-AsyncOperation` headers then it MUST have the property `x-ms-long-running-operation` set to `true`.
+
+## Description
+
+If an operation's (PUT/POST/PATCH/DELETE) responses have `Location` or `Azure-AsyncOperation` headers then it MUST have the property `x-ms-long-running-operation` set to `true`.
+
+## How to fix the violation
+
+You can do either of the following:
+- Set the property `x-ms-long-running-operation` to `true`.
+- If `x-ms-long-running-operation` is set to `false` then remove `Location` and `Azure-AsyncOperation` headers from the response.
+
+### Valid Examples
+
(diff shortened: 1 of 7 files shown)


docs/xms-long-running-operation-property.md
```diff
@@ -0,0 +1,67 @@
...
+
+## Description
+
+If an operation's (PUT/POST/PATCH/DELETE) responses have `Location` or `Azure-AsyncOperation` headers then it MUST have the property `x-ms-long-running-operation` set to `true`.
+
+## How to fix the violation
+
+You can do either of the following:
+- Set the property `x-ms-long-running-operation` to `true`.
+- If `x-ms-long-running-operation` is set to `false` then remove `Location` and `Azure-AsyncOperation` headers from the response.
```
rkmanda, line 27: >If `x-ms-long-running-operation` is set to `false` then remove `Location` and `Azure-AsyncOperation` headers from the response. [](http://example.com/codeflow?start=2&length=127)

If the operation is not intended to be a long running operation, `x-ms-long-running-operation` MUST be set to `false` and the `Location` and `Azure-AsyncOperation` headers MUST not be specified in the response.

docs/xms-long-running-operation-property.md
```diff
@@ -0,0 +1,67 @@
...
+If an operation's (PUT/POST/PATCH/DELETE) responses have `Location` or `Azure-AsyncOperation` headers then it MUST have the property `x-ms-long-running-operation` set to `true`.
+
+## Description
+
+If an operation's (PUT/POST/PATCH/DELETE) responses have `Location` or `Azure-AsyncOperation` headers then it MUST have the property `x-ms-long-running-operation` set to `true`.
+
+## How to fix the violation
+
+You can do either of the following:
+- Set the property `x-ms-long-running-operation` to `true`.
```
rkmanda, line 26: > Set the property `x-ms-long-running-operation` to `true`. [](http://example.com/codeflow?start=1&length=58)

If the operation is intended to be a long running operation, set the property `x-ms-long-running-operation` to `true`.

rkmanda (approved):
:shipit:

docs/xms-long-running-operation-property.md
tejaswiMinnu: Added

docs/xms-long-running-operation-property.md
tejaswiMinnu: Added

PR #587, opened 2023-09-08
Add test cases for path-for-resource-action linter rule

AkhilaIlla:
RPC-POST-V1-07 rule is already covered by PathForResourceAction linter rule
Added more tests to verify if the rule satisfies all the cases in RPC-POST-V1-07


rkmanda (approved):
:shipit:

PR #595, opened 2023-09-21
Print out details of AggregateError originating from Spectral

konrad-jamrozik:
Fixes:
- https://github.com/Azure/azure-sdk-tools/issues/6856

See [this comment](https://github.com/Azure/azure-sdk-tools/issues/6856#issuecomment-1728908191) for details.

### DIFF
--- a/packages/azure-openapi-validator/autorest/src/spectral-plugin-func.ts
+++ b/packages/azure-openapi-validator/autorest/src/spectral-plugin-func.ts
@@ -80,10 +80,7 @@ export async function spectralPluginFunc(initiator: IAutoRestPluginInitiator): P
       const normalizedFile = file.startsWith("file:///") ? fileURLToPath(file) : file
       await runSpectral(openapiDefinitionObject, normalizedFile, initiator.Message.bind(initiator), spectral)
     } catch (e) {
-      initiator.Message({
-        Channel: "fatal",
-        Text: `spectralPluginFunc: Failed validating: '${file}', error encountered: ` + e,
-      })
+      catchSpectralRunErrors(file, e, initiator)
     }
   }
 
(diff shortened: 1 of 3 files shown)


packages/azure-openapi-validator/autorest/src/spectral-plugin-func.ts
```diff
@@ -233,3 +230,29 @@ export async function getRuleSet(openapiType: OpenApiTypes): Promise<[ruleset: R
...
   }
 }
+
+/** Spectral (from "@stoplight/spectral-core") may throw https://www.npmjs.com/package/es-aggregate-error
+ * If so, we print out all the constituent errors.
+ * For additional context, see: https://github.com/Azure/azure-sdk-tools/issues/6856
+ */
+function catchSpectralRunErrors(file: string, e: any, initiator: any) {
+  // Initialize an array to collect error messages
+  const errorMessages: string[] = [e]
```
TimLovellSmith, line 240: does initiating it with 'e' make sense when e (typed `any`) is not actually a string?

mentat9: Also likely to format as weird or misleading...? I don't know what e coerced to string as the first "error message" looks like, but even if it looks like an error message, won't it be duplicate of messages that are added after?

konrad-jamrozik: @TimLovellSmith, @mentat9, example output:

e is `AggregateError`:

```
fatal   | spectralPluginFunc: Failed validating: 'file:///C:/Users/spawa/repos/specs/specification/app/resource-manager/Microsoft.App/preview/2023-08-01-preview/Usages.json'. Errors encountered:
AggregateError: Error running Nimma
Error 1: $[paths,'x-ms-paths'].* threw: ErrorWithCause("Function \"patchPropertiesCorrespondToPutProperties\" threw an exception: Cannot read properties of undefined (reading 'length')")
Error 2: $[paths,'x-ms-paths'].* threw: ErrorWithCause("Function \"patchPropertiesCorrespondToPutProperties\" threw an exception: Cannot read properties of undefined (reading 'length')")
```
image

e is not `AggregateError`: I am unsure if this case can even happen.

My take is to leave it as it is for now. I suspect treating `e` as string will always lead to good output. I will monitor the the output once we deploy it and adjust if necessary. If anything, perhaps the severity should be changed from `fatal` to `error`, but maybe not.

mentat9 (approved):

PR #599, opened 2023-09-25
Update deploy_to_prod_section in readMe file

AkhilaIlla:
Update deploy_to_prod_section in contibuting.md file to make instructions more clear

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -247,12 +247,11 @@ If you want your changes to be deployed to [production pipeline](https://dev.azu
 and hence Production LintDiff, you need to do the following:
 
 - In the PR with your changes increase the version number of the packages you changed.
-  - [Here](https://github.com/Azure/azure-openapi-validator/pull/506/files#diff-cad0ec93b3ac24499b20ae58530a4c3e7f369bde5ba1250dea8cad8201e75c30)
-    is an example version increase for the ruleset.
-  - And [here](https://github.com/Azure/azure-openapi-validator/pull/506/files#diff-359645f2d25015199598e139bc9b03c9fec5d5b1a4a0ae1f1e4f7a651675e6bf)
-    for changes made to the  AutoRest extension.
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -247,12 +247,11 @@ If you want your changes to be deployed to [production pipeline](https://dev.azu
 and hence Production LintDiff, you need to do the following:
 
 - In the PR with your changes increase the version number of the packages you changed.
-  - [Here](https://github.com/Azure/azure-openapi-validator/pull/506/files#diff-cad0ec93b3ac24499b20ae58530a4c3e7f369bde5ba1250dea8cad8201e75c30)
-    is an example version increase for the ruleset.
-  - And [here](https://github.com/Azure/azure-openapi-validator/pull/506/files#diff-359645f2d25015199598e139bc9b03c9fec5d5b1a4a0ae1f1e4f7a651675e6bf)
-    for changes made to the  AutoRest extension.
+  - Bump the version in **package.json** under ruleset folder only if you made changes to Ruleset extension(i.e, changes made are under ruleset folder). Example [Here](https://github.com/Azure/azure-openapi-validator/pull/506/files#diff-cad0ec93b3ac24499b20ae58530a4c3e7f369bde5ba1250dea8cad8201e75c30).
```
konrad-jamrozik, line 250: Few remarks:
- not `extension` but `package`.
- missing space before `(`
- `made are` -> `made`
- there is no `ruleset` folder. Plus, it is good to provide a full path relative to repo root, and also hyperlink to it. So something like "(i.e. changes made under the [packages/rulesets](https://github.com/Azure/azure-openapi-validator/tree/main/packages/rulesets) folder)"

Also, the problem with this guidance is that it is very shallow: once one understands that `package.json` applies to the directory in which it is located, this becomes self-evident. So I would like for the documentation to point that out, educating the reader how things work so they have a better understanding what is going on. 

Something like:

- Note that a package encompasses a directory in which the `package.json` defining it is located. Hence, for example, if you made any changes to files under [packages/rulesets](https://github.com/Azure/azure-openapi-validator/tree/main/packages/rulesets), then you need to update [the version in packages/rulesets.package.json](https://github.com/Azure/azure-openapi-validator/blob/main/packages/rulesets/package.json#L3). The same applies to all the other directories/packages in [packages](https://github.com/Azure/azure-openapi-validator/tree/main/packages) directory.

AkhilaIlla: Thanks Konrad!
makes more sense and the links are helpful as well.

konrad-jamrozik (approved):

PR #596, opened 2023-09-21
Update PatchPropertiesCorrespondToPutProperties to run in stagingOnly

AkhilaIlla:
The LintDiff pipeline is failing with 

```
{
  "pluginName": "spectral",
  "extensionName": "@microsoft.azure/openapi-validator",
  "level": "fatal",
  "message": "spectralPluginFunc: Failed validating: 'file:///mnt/vss/_work/1/azure-rest-api-specs pr/specification/resources/resource-manager/Microsoft.Resources/moboBrokers/preview/2023-06-01-preview/moboBrokers.json', error encountered: Error: Error running Nimma"
}
```

https://teams.microsoft.com/l/message/19:4f661242c446452e895359cc3ef45125@thread.tacv2/1695277014925?tenantId=72f988bf-86f1-41af-91ab-2d7cd011db47&groupId=0cab4ce9-7691-42ae-82e3-460d4346a710&parentMessageId=1692914350538&teamName=ARM%20API%20Reviewers&channelName=ARM%20linter%20rules&createdTime=1695277014925
and hence the faulty rule is updated to run only in staging

PR #603, opened 2023-09-29
Bugfix: actually disable disabled rules

konrad-jamrozik:
A follow-up bugfix to:
- https://github.com/Azure/azure-openapi-validator/pull/595
- https://github.com/Azure/azure-openapi-validator/pull/596

See the comment in the diff for details.

Companion PR:
- https://devdiv.visualstudio.com/DevDiv/_git/openapi-alps/pullrequest/501040

### DIFF
--- a/packages/azure-openapi-validator/autorest/src/spectral-plugin-func.ts
+++ b/packages/azure-openapi-validator/autorest/src/spectral-plugin-func.ts
@@ -116,6 +116,18 @@ function ifNotStagingRunThenDisableStagingOnlyRules(
         // Resulting in value of -1, as defined here:
         // https://github.com/stoplightio/spectral/blob/44c40e2b7c9ea6222054da8700049b0307cc5f8b/packages/ruleset-migrator/src/transformers/rules.ts#L39
         rule.severity = "off"
+        // We must disable the rule to ensure it is not run at all. Spectral source for that is in [1].
+        // Otherwise, if it throws due to a bug in rule implementation, it will result in fatal error.
+        // Example of how a rule can thrown is given in [2].
+        // Example where we disable a rule from running in production so it doesn't throw is given in [3].
+        // Without this line, the fix in [3] doesn't help. For a proof of that, see [4] and its log, [5],
+        // showing that a rule that is supposed to run in stagingOnly, still throws in production.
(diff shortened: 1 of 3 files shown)


packages/azure-openapi-validator/autorest/src/spectral-plugin-func.ts
```diff
@@ -116,6 +116,18 @@ function ifNotStagingRunThenDisableStagingOnlyRules(
...
+        // Example of how a rule can thrown is given in [2].
+        // Example where we disable a rule from running in production so it doesn't throw is given in [3].
+        // Without this line, the fix in [3] doesn't help. For a proof of that, see [4] and its log, [5],
+        // showing that a rule that is supposed to run in stagingOnly, still throws in production.
+        // [1] https://github.com/stoplightio/spectral/blob/6d0991572f185ce5cf4031dc1d8eb4035b5eaf1d/packages/core/src/runner/runner.ts#L39
+        // [2] https://github.com/Azure/azure-openapi-validator/pull/595
+        // [3] https://github.com/Azure/azure-openapi-validator/pull/596
+        // [4] https://github.com/Azure/azure-rest-api-specs/pull/26024
+        // [5] https://dev.azure.com/azure-sdk/internal/_build/results?buildId=3125612&view=logs&j=0574a2a6-2d0a-5ec6-40e4-4c6e2f70bea2&t=80c3e782-49f0-5d1c-70dd-cbee57bdd0c7&l=252
+        rule.enabled = false
```
AkhilaIlla, line 130: enabled would be marked as false for all the staging rules from now on with your fix?
and from [1] looks like all the enabled rules are considered to run, then how/why the other staging rules didnt show up in prod earlier?

AkhilaIlla (approved):

packages/azure-openapi-validator/autorest/src/spectral-plugin-func.ts
konrad-jamrozik: If you look [at the function in which this change is made](https://github.com/Azure/azure-openapi-validator/blob/554381e97808cf6822c8b596a20e0ab72790bd38/packages/azure-openapi-validator/autorest/src/spectral-plugin-func.ts#L93-L134) it should become clear - this logic executes only if rule has been disabled. Hence a rule becomes disabled only if we are running production LintDiff and rule is marked as `stagingOnly`.

PR #606, opened 2023-10-06
Tejaswis/standardize md files

tejaswiMinnu:
standardize .md file format

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -448,7 +448,7 @@ ARM OpenAPI(swagger) specs
 
 ## Related ARM Guideline Code
 
--<add code, example: "-RPC-Uri-V1-04">
+-<add RPC code, example: "-RPC-Uri-V1-04">
 
 ## Description
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -436,13 +436,35 @@ and follow below steps to add a rule for azure rest api specs.
 - add a doc for the rule in 'docs', the rule doc file name is following kebab-case, contains following content:
 
 ```md
-# RuleName
+# <add RuleName>
+
+## Category
+
+ARM Error
+
+## Applies to
+
+ARM OpenAPI(swagger) specs
+
+## Related ARM Guideline Code
+
+-<add code, example: "-RPC-Uri-V1-04">
 
 ## Description
 
 <add the description for the rule.>
+
 ## How to fix
+
 <describe how to fix the violations.>
+
+## Good Examples
+
+<add valid json snippet>
+
+## Bad Examples
+
+<add invalid json snippet>
 ```
 
 ## Native rule
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -456,15 +456,28 @@ and follow below steps to add a rule for azure rest api specs.
 
 ## Category
 
+<add one of the following>
+SDK Error
+SDK Warning
+RPaaS Error
+RPaaS Warning
 ARM Error
+ARM Warning
+Data Plane Error
+Data Plane Warning
+ARM & Data Plane Error
+ARM & Data Plane Warning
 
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -436,13 +436,35 @@ and follow below steps to add a rule for azure rest api specs.
 - add a doc for the rule in 'docs', the rule doc file name is following kebab-case, contains following content:
 
 ```md
-# RuleName
+# <add RuleName>
+
+## Category
+
+ARM Error
```
AkhilaIlla, line 443: this needs to be generic place holder
could be ARM error or datat place or both right?

CONTRIBUTING.md
```diff
@@ -436,13 +436,35 @@ and follow below steps to add a rule for azure rest api specs.
...
+
+## Applies to
+
+ARM OpenAPI(swagger) specs
```
AkhilaIlla, line 447: same

CONTRIBUTING.md
```diff
@@ -436,13 +436,35 @@ and follow below steps to add a rule for azure rest api specs.
...
 
 ## Description
 
 <add the description for the rule.>
+
 ## How to fix
+
 <describe how to fix the violations.>
+
+## Good Examples
```
AkhilaIlla, line 461: I see a lot of existing .md files have "output message" section too. Do we plan to remove it?

CONTRIBUTING.md
```diff
@@ -436,13 +436,35 @@ and follow below steps to add a rule for azure rest api specs.
...
+
+## Category
+
+ARM Error
+
+## Applies to
+
+ARM OpenAPI(swagger) specs
+
+## Related ARM Guideline Code
```
bdefoy, line 449: On the same topic as Akhila, this section should be optional because data plane linter rules would not have ARM codes

CONTRIBUTING.md
tejaswiMinnu: We can update this to "ARM/SDK Warning", "ARM/SDK Error" but not considered Data Plane rules. We are checking linter rules only on "azure-rest-api-specs", so I mentioned only "ARM OpenAPI(swagger) specs" 
@rkmanda can you please confirm if I need to update these standards to satisfy Data Plane rules as well?

CONTRIBUTING.md
tejaswiMinnu: The whole section is not mandate, we just want to standardize the format to be in sync with all other .md files

CONTRIBUTING.md
tejaswiMinnu: Yes, output message is same as error message so felt redundant adding in both the places.

CONTRIBUTING.md
rkmanda: lets provide all possible values here

CONTRIBUTING.md
rkmanda: yup lets specify all possible values here

CONTRIBUTING.md
rkmanda: you can mention explicitly that this is mandatory when authoring an ARM guideline

<add RPC code for ARM guidelines, example: "-RPC-Uri-V1-04". This is not required for data plane guidelines>

rkmanda (changes requested):
:clock1:

CONTRIBUTING.md
tejaswiMinnu: Added all values

CONTRIBUTING.md
tejaswiMinnu: Added

CONTRIBUTING.md
tejaswiMinnu: Changed

rkmanda (approved):
:shipit:

PR #612, opened 2023-10-11
Fix ResourceNameRestriction from throwing FATAL error and version bump

AkhilaIlla:
ResourceNameRestriction is throwing FATAL error -   https://dev.azure.com/azure-sdk/internal/_build/results?buildId=3154966&view=logs&j=0574a2a6-2d0a-5ec6-40e4-4c6e2f70bea2&t=80c3e782-49f0-5d1c-70dd-cbee57bdd0c7&l=202035

This PR fixes the issue 

More info here:
https://teams.microsoft.com/l/message/19:4f661242c446452e895359cc3ef45125@thread.tacv2/1696815957907?tenantId=72f988bf-86f1-41af-91ab-2d7cd011db47&groupId=0cab4ce9-7691-42ae-82e3-460d4346a710&parentMessageId=1696815957907&teamName=ARM%20API%20Reviewers&channelName=ARM%20linter%20rules&createdTime=1696815957907

### DIFF
--- a/packages/rulesets/src/spectral/functions/resource-name-restriction.ts
+++ b/packages/rulesets/src/spectral/functions/resource-name-restriction.ts
@@ -10,14 +10,22 @@ export const resourceNameRestriction = (paths: any, _opts: any, ctx: any) => {
 
   function getPathParameter(pathItem: any, paramName: string) {
     let parameters: any[] = []
-    const method = Object.keys(pathItem).find((k) => k !== "parameters")
-    if (method) {
+    // Get all the methods defined 
+    const methods = Object.keys(pathItem).filter((k) => k !== "parameters")
+
+    for (const method of methods) {
       const operationParameters = pathItem[method].parameters
-      parameters = parameters.concat(operationParameters)
+      // Add parameters if exists
+      if (operationParameters) {
+        parameters = parameters.concat(operationParameters)
+      }
     }
+    // Concatenate path parameters if exists
     if (pathItem.parameters) {
       parameters = parameters.concat(pathItem.parameters)
     }
+
+    // Filter down to desired path parameter
     return parameters.find((p) => p.in === "path" && p.name === paramName)
   }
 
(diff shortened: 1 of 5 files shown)


packages/rulesets/src/spectral/functions/resource-name-restriction.ts
```diff
@@ -10,14 +10,22 @@ export const resourceNameRestriction = (paths: any, _opts: any, ctx: any) => {
...
-    if (method) {
+    // Get all the methods defined 
+    const methods = Object.keys(pathItem).filter((k) => k !== "parameters")
+
+    for (const method of methods) {
       const operationParameters = pathItem[method].parameters
-      parameters = parameters.concat(operationParameters)
+      // Add parameters if exists
+      if (operationParameters) {
+        parameters = parameters.concat(operationParameters)
```
weshaggard, line 20: Are there any cases where a parameter might be null itself? Or are there any cases where a parameter might not have an "in" or "name" property defined which would case the return statement to fail?

mikekistler:
I think I found and fixed this same problem.  See #613.

packages/rulesets/src/spectral/functions/resource-name-restriction.ts
mikekistler, line 20: I don't think it is necessary or desirable to iterate through all the methods - every method within the path will have exactly the same path parameters so just picking the first one is all that's needed.

packages/rulesets/src/spectral/test/resource-names-restriction.test.ts
```diff
@@ -93,3 +93,427 @@ test("ResourceNameRestriction should find no errors", () => {
     expect(results.length).toBe(0)
   })
 })
+
+test("ResourceNameRestriction should find errors for multiple methods and undefined parameters section", () => {
```
mikekistler, line 97: I think this huge new test is unnecessary. The only thing that needs to verified is that the function does not crash for an operation with no `parameters`.

mikekistler (commented):
I think the fix and test in #613 are more simple and concise.

AkhilaIlla:
This issue is taken care in https://github.com/Azure/azure-openapi-validator/pull/613

PR #613, opened 2023-10-11
Fix resourceNameRestriction to handle missing operation parameters

mikekistler:
This PR makes a little fix to the resourceNameRestriction function to correctly handle an operation that does not define any parameters.

This problem was exposed by [PR 25997](https://github.com/Azure/azure-rest-api-specs/pull/25977).

Test included.


rkmanda (approved):
:shipit:

weshaggard (approved):
You will need to also bump the version of the ruleset packages similar to what @AkhilaIlla did at https://github.com/Azure/azure-openapi-validator/pull/612/files#diff-cad0ec93b3ac24499b20ae58530a4c3e7f369bde5ba1250dea8cad8201e75c30

PR #614, opened 2023-10-23
Organize and add to roll out section of contributing doc

bdefoy:
- Add information about communications with stakeholders for new rules
- Reorganize the `How to add and roll out new linter rules` section


rkmanda (approved):
:shipit:

PR #622, opened 2023-11-14
[LintDiff] Disable select rules for Typespec.

### LINKED ISSUE #618: [LintDiff] Need separate ruleset for TypeSpec-generated Swaggers
We need the following capabilities in LintDiff:
1. Creation of a distinct ruleset for TypeSpec-generated Swaggers, which is a subset of the normal ruleset for hand-authored Swaggers. There should never be a rule in the TypeSpec ruleset that is not in the normal ruleset. The goal will be to shrink the number of rules in this ruleset until it is empty, and no LintDiff rules run against TypeSpec-generated Swaggers.
2. We need a flag or other parameter to tell LintDiff which ruleset to apply. 

The mechanism will allow us to accomplish the following goals:
1. Avoid a LintDiff failure when the corresponding TypeSpec rule has been purposefully suppressed. Authors should not have to suppress the same thing twice.
2. Allows us to use LintDiff as a backstop for catching issues when an equivalent check does not exist in TypeSpec. As we add equivalent checks to TypeSpec, or identify rules that aren't relevant to TypeSpec-generated Swagger, we will remove them from this ruleset. Eventually, this ruleset will be small or empty.

tjprescott:
Closes Azure/azure-sdk-tools#7732.


packages/azure-openapi-validator/autorest/src/spectral-plugin-func.ts
```diff
@@ -36,7 +36,8 @@ export async function spectralPluginFunc(initiator: IAutoRestPluginInitiator): P
   }
 
   const resolvedOpenapiType: OpenApiTypes = getOpenapiType(openapiType)
-  const [ruleset, namesOfRulesInStagingOnly]: [Ruleset, string[]] = await getRuleSet(resolvedOpenapiType)
+  // FIXME: This should be set by the CI, not hardcoded
```
mikeharder, line 39: I think some component should look for the `x-typespec-generated` flag in swagger to decide whether to enable this.

konrad-jamrozik: Can you please elaborate (in code or comment) what is the `this` here?

packages/azure-openapi-validator/autorest/src/spectral-plugin-func.ts
```diff
@@ -36,7 +36,8 @@ export async function spectralPluginFunc(initiator: IAutoRestPluginInitiator): P
...
+  const [ruleset, namesOfRulesInStagingOnly]: [Ruleset, string[]] = await getRuleSet(resolvedOpenapiType, true)
```
konrad-jamrozik, line 40: Let's clarify what is this `true` at the end ? Either with appropriately named variable or a comment.

packages/azure-openapi-validator/autorest/src/spectral-plugin-func.ts
```diff
@@ -228,10 +232,29 @@ export async function getRuleSet(openapiType: OpenApiTypes): Promise<[ruleset: R
...
   const namesOfRulesInStagingOnly: string[] = processRulesInStagingOnly(ruleset.rules)
 
   return [new Ruleset(ruleset, { severity: "recommended" }), namesOfRulesInStagingOnly]
 
+  /**
+   * Filters out properties from a ruleset that are deemed irrelevant to Typespec-generated Swagger.
+   * @param ruleset The ruleset to be filtered.
+   * @returns The ruleset with rules that are not applicable to Typespec removed.
+   */
+  function processRulesApplicableToTypespec(ruleset: any) {
```
konrad-jamrozik, line 248: suggest to rename the function to `removeRulesIgnoredForTypespec`

packages/azure-openapi-validator/autorest/src/spectral-plugin-func.ts
```diff
@@ -212,7 +213,10 @@ async function runSpectral(doc: any, filePath: string, sendMessage: (m: Message)
   return mergedResults
 }
 
-export async function getRuleSet(openapiType: OpenApiTypes): Promise<[ruleset: Ruleset, namesOfRulesInStagingOnly: string[]]> {
+export async function getRuleSet(
+  openapiType: OpenApiTypes,
+  filterForTypespec: boolean
```
konrad-jamrozik, line 218: suggest to rename `filterForTypespec` to `removeRulesIgnoredByTypespec`

packages/rulesets/generated/spectral/az-arm.js
```diff
@@ -843,6 +843,7 @@ const ruleset$1 = {
             description: "Operation should have a summary or description.",
             message: "Operation should have a summary or description.",
             severity: "warn",
+            ignoreForTypespec: true,
```
konrad-jamrozik, line 846: Ideally each such line should have a comment explaining why this is here. Either provide a URL showing where is the validation that replaces this LintDiff validation, or provide a URL to the issue to add such validation (and ideally that issue would say why it is OK to disable this).

If it is not possible to add comments to these ruleset files (not sure if it is) then maybe we could add extra ignored field like `ignoreForTypespecReason`.

packages/azure-openapi-validator/autorest/src/spectral-plugin-func.ts
konrad-jamrozik, line 39: Can you elaborate what is the `this` here?

packages/rulesets/generated/spectral/az-arm.js
mikeharder: I would also feel more confident if we could add tests for each rule, where we take a small bit of typespec, generate swagger, then run both tools and make sure the typespec rule covers.

packages/azure-openapi-validator/autorest/src/spectral-plugin-func.ts
mikeharder: The second parameter is hardcoded to `true`, it should be set to `does swagger contain x-typespec-generated`.

packages/azure-openapi-validator/autorest/src/spectral-plugin-func.ts
tjprescott: That's the same name you suggested for the method, so it causes a name conflict. I used `isTypespecGenerated` instead

packages/azure-openapi-validator/autorest/src/spectral-plugin-func.ts
konrad-jamrozik: Yeah, I think it would be cool if the comment in the code spells that out.

packages/rulesets/generated/spectral/az-dataplane.js
```diff
@@ -1942,6 +1946,7 @@ const ruleset = {
     rules: {
         AdditionalPropertiesAndProperties: {
             description: "Don't specify additionalProperties as a sibling of properties.",
+            ignoreForTypespec: true,
```
johanste, line 1949: I'm not sure if we have a complete guard against this happening in TSP.

johanste:
With the possible exception for the additional properties + properties rule (added comment), I see no significant harm even if we accidentally let some specs through here. If we can reduce the noise, I would be very much in favor of getting this in.

packages/rulesets/generated/spectral/az-dataplane.js
timotheeguerin: we do `@azure-tools/typespec-azure-core/bad-record` https://azure.github.io/typespec-azure/playground?c=LyoqIFdpdGggZG9jICovCm1vZGVsIEZvbyBpcyBSZWNvcmQ8c3RyaW5nPiB7CiAg0DAgIG5hbWU6IMYkOwp9Cg%3D%3D&e=%40azure-tools%2Ftypespec-autorest&options=%7B%22linterRuleSet%22%3A%7B%22extends%22%3A%5B%22%40azure-tools%2Ftypespec-azure-core%2Fall%22%5D%7D%7D

johanste: Gah, I was using a [stale playground link](https://cadlplayground.z22.web.core.windows.net/cadl-azure/?c=bW9kZWwgRG9lc0l0V29yayBleHRlbmRzIFJlY29yZDxzdHJpbmc%2BIHsKICBiYXI6IMYROwp9&e=%40azure-tools%2Ftypespec-autorest&options=%7B%7D) when I tested.

timotheeguerin: That links works too but the linters are not enabled by default. You have to select them

konrad-jamrozik:
Closing as subsumed by:
- https://github.com/Azure/azure-openapi-validator/pull/663

PR #623, opened 2023-11-14
Fixes the issue with PatchPropertiesCorrespondToPutProperties

tejaswiMinnu:
The PatchPropertiesCorrespondToPutProperties rule should verify whether a given path contains both PUT and PATCH calls. If either of this call is missing, the lint check should raise a fatal error. The comparison of properties between these two calls becomes problematic when one of them (in this case, PATCH) is undefined. Consequently, applying a filter to retrieve properties from an undefined variable is not feasible.

PR which failed with Fatal Errors: https://github.com/Azure/azure-rest-api-specs/pull/26281/checks?check_run_id=17848084756

### DIFF
--- a/common/changes/@microsoft.azure/openapi-validator-rulesets/fps_patchPropertiesCorrespondToPutProperties_2023-11-14-22-01.json
+++ b/common/changes/@microsoft.azure/openapi-validator-rulesets/fps_patchPropertiesCorrespondToPutProperties_2023-11-14-22-01.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@microsoft.azure/openapi-validator-rulesets",
+      "comment": "Fix an issue where PatchPropertiesCorrespondToPutProperties rule checks if the path has both PUT and PATCH calls.",
+      "type": "patch"
+    }
+  ],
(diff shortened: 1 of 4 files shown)


AkhilaIlla (approved):

bdefoy (approved):

rkmanda (approved):
:shipit:

common/changes/@microsoft.azure/openapi-validator-rulesets/fps_patchPropertiesCorrespondToPutProperties_2023-11-14-22-01.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@microsoft.azure/openapi-validator-rulesets",
+      "comment": "Fix an issue where PatchPropertiesCorrespondToPutProperties rule checks if the path has both PUT and PATCH calls.",
```
bdefoy, line 5:
```suggestion
      "comment": "Fix an issue where the PatchPropertiesCorrespondToPutProperties rule would fail if either PATCH or PUT operations were undefined.",
```

PR #624, opened 2023-11-15
Fix fatal error caused by undefined parameters in ParametersInPointGet

AkhilaIlla:
Undefined params cause the rule to fail with fatal error-https://github.com/Azure/azure-rest-api-specs/pull/26218/checks?check_run_id=18133490258
This PR fixes the issue


rkmanda (approved):
:shipit:

bdefoy (approved):

PR #631, opened 2023-12-05
Fix trackedExtensionResourcesAreNotAllowed from flagging false positives

AkhilaIlla:
Found trackedExtensionResourcesAreNotAllowed wrongly flags
```/subscriptions/{subscriptionId}/resourcegroups/{resourceGroupName}/providers/Microsoft.Kubernetes/connectedClusters/{clusterName}``` kind of paths

But, Isn't a extension resource. 

This PR fixes such issues. Fixed the test to catch this issue too.


rkmanda:
>         get: {

This path will still match the new regex and get flagged as an extension resource. The expectation is that it should not. 

---
Refers to: packages/rulesets/src/spectral/test/tracked-extension-resources-are-not-allowed.test.ts:734 in dfc71de. [](commit_id = dfc71de3715f39921ce49ca7d745b9bf979d5617, deletion_comment = False)

rkmanda:
>         get: {

The reason why it might not be getting flagged is possibly because the definition is not getting resolved correctly? Its just a hypothesis but pl doublecheck this.

---
In reply to: [1846461736](https://github.com/Azure/azure-openapi-validator/pull/631#issuecomment-1846461736) [](http://example.com/codeflow?ancestors=1846461736)

---
Refers to: packages/rulesets/src/spectral/test/tracked-extension-resources-are-not-allowed.test.ts:734 in dfc71de. [](commit_id = dfc71de3715f39921ce49ca7d745b9bf979d5617, deletion_comment = False)

rkmanda:
>       DeviceProvisioningState: {

Model name sin Definitions need to be enclosed in double quotes.. so this needs to be "DeviceProvisioningState"

---
Refers to: packages/rulesets/src/spectral/test/tracked-extension-resources-are-not-allowed.test.ts:834 in dfc71de. [](commit_id = dfc71de3715f39921ce49ca7d745b9bf979d5617, deletion_comment = False)

AkhilaIlla:
>         get: {

These are tracked non-extension resources & shouldnt be flagged?

and the test covers it?

Are you trying to say something else?

---
In reply to: [1846462992](https://github.com/Azure/azure-openapi-validator/pull/631#issuecomment-1846462992) [](http://example.com/codeflow?ancestors=1846462992,1846461736)

---
Refers to: packages/rulesets/src/spectral/test/tracked-extension-resources-are-not-allowed.test.ts:734 in dfc71de. [](commit_id = dfc71de3715f39921ce49ca7d745b9bf979d5617, deletion_comment = False)

rkmanda (approved):
:shipit:

PR #634, opened 2023-12-08
Change the rule AllTrackedResourcesMustHaveDelete to flag an error even when a Put operation is not specified

rkmanda:
The rule AllTrackedResourcesMustHaveDelete flags an error when a Delete operation is not specified for a Tracked resource only when a Put is specified. This rule must flag an error irrespective of whether a Put is specified or not for a tracked resource. This change ensures the same.

### DIFF
--- a/docs/all-proxy-resources-should-have-delete.md
+++ b/docs/all-proxy-resources-should-have-delete.md
@@ -12,22 +12,299 @@ ARM OpenAPI(swagger) specs
 
 - RPC-Delete-V1-05
 
-## Output Message
-
-The resource {resourceName} does not have a corresponding delete operation.
-
 ## Description
 
-All proxy resources SHOULD support delete.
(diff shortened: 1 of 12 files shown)


docs/all-proxy-resources-should-have-delete.md
```diff
@@ -12,22 +12,299 @@ ARM OpenAPI(swagger) specs
...
-## Output Message
+## Description
 
-The resource {resourceName} does not have a corresponding delete operation.
+All proxy resources that have a put operation defined should also support a delete operation. This makes the APIs intuitive for customers as they would expect every resource that they have the ability to create to also be deletable. There are rare sceanrios where a customer is allowed to create and replace a resource but never be able to delete them. In such cases it is ok to not implement a delete but such cases are very rare and should be avoided. 
 
-## Description
+## How to fix
+
+Add the delete operation for the proxy resource.
```
AkhilaIlla, line 21: an we add "If PUT exists" to the end as DEL isnt mandatory if PUT isnt defined

AkhilaIlla (approved):

docs/all-proxy-resources-should-have-delete.md
rkmanda: The assumption is that the PUT doesnt exist as thats when the rule flags an error but I can make it explicit.

PR #635, opened 2023-12-09
Fix client flatten documentation and add tests

rkmanda:
An issue was raised indicating that the linter rule was raising a false alarm in this PR

https://github.com/Azure/azure-rest-api-specs/pull/23563/checks?check_run_id=12809910224

On investigating this it was found that the linter rule was working correctly. The issue was that the type:object declaration was missing where it should have been present.

This change is to update the documentation to call out explicitly what the good and bad patterns are. This change also adds more tests

### DIFF
--- a/docs/avoid-nested-properties.md
+++ b/docs/avoid-nested-properties.md
@@ -20,6 +20,101 @@ Nested properties can result into bad user experience especially when creating r
 
 Overly nested properties (especially required ones) can result into a non optimal user experience.
 
-## How to fix the violation
+## How to fix
 
 Either eliminate nesting or use the `x-ms-client-flatten` property for a better user experience. More details about the extension can be found [here](https://github.com/Azure/autorest/blob/main/docs/extensions/readme.md#x-ms-client-flatten).
+
+## Bad examples
+
+## Bad example 1
+Deeply nested properties exist and the x-ms-client-flatten extension is not specified
+
+```json
+    definitions: {
+      externalDocs: {
+        type: "object",
+        properties: {
+          tags: {
+            type: "object",
+            additionalProperties: {
+              type: "string",
+            },
+            description: "Resource tags.",
+          },
+          properties: {
+            type: "object",
+            properties: {
+              timeOfDayUTC: {
+                type: "string",
+                pattern: "^([0-9]|0[0-9]|1[0-9]|2[0-3]):[0-5][0-9]$",
+                description: "The time of day (in UTC) to start the maintenance window.",
+              },
+            },
+          },
+        },
+      },
+    },
+```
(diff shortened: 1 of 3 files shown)


docs/avoid-nested-properties.md
```diff
@@ -20,6 +20,101 @@ Nested properties can result into bad user experience especially when creating r
...
+            additionalProperties: {
+              type: "string",
+            },
+            description: "Resource tags.",
+          },
+          properties: {
+            type: "object",
+            // x-ms-client-flatten should have been specified here
+            properties: {
+              "x-ms-client-flatten": true,
```
ravimeda, line 78: Minor - provide an example to show the first option i.e., remove the nesting.

ravimeda (approved):

AkhilaIlla:
> Consider using x-ms-client-flatten to provide a better end user experience

As we are fixing the doc, can we remove this section?
To match the standard doc format

---
Refers to: docs/avoid-nested-properties.md:13 in 92f212e. [](commit_id = 92f212e877009121049837d48c6f2d7bb7ef3ec2, deletion_comment = False)

packages/rulesets/src/spectral/test/avoid-nested-properties.test.ts
```diff
@@ -64,3 +73,79 @@ test("AvoidNestedProperties should find no errors", () => {
...
   })
 })
+
+test("AvoidNestedProperties should find errors when the extension is specified outside the complex type definition", () => {
+  const myOpenApiDocument = {
+    swagger: "2.0",
+    info: {
+      description:
+        "Use these APIs to manage Azure Websites resources through the Azure Resource Manager. All task operations conform to the HTTP/1.1 protocol specification and each operation returns an x-ms-request-id header that can be used to obtain information about the request. You must make sure that requests made to these resources are secure. For more information, see https://msdn.microsoft.com/en-us/library/azure/dn790557.aspx.",
+    },
```
AkhilaIlla, line 83: As its just a test, not sure if we'd need this very descriptive description

AkhilaIlla (approved):

packages/rulesets/src/spectral/test/avoid-nested-properties.test.ts
rkmanda: hmm doesnt hurt to have a meaningful description

docs/avoid-nested-properties.md
rkmanda: Actually in the examples used so far, it is not possible to remove the deep nesting. and usually its not a practical solution. I will remove the part which says remove deep nesting

PR #639, opened 2023-12-15
Update doc for ProvisioningStateMustBeReadOnly

AkhilaIlla:
Folks complain ProvisioningStateMustBeReadOnly flagging false positives, when it isnt. This PR improves the doc to include all the good and bad examples and make docs more clear

### DIFF
--- a/docs/provisioning-state-must-be-read-only.md
+++ b/docs/provisioning-state-must-be-read-only.md
@@ -8,6 +8,10 @@ ARM Error
 
 ARM OpenAPI(swagger) specs
 
+## Related ARM Guideline Code
+
+- RPC-Async-V1-16
+
 ## Description
 
 This is a rule introduced to validate if provisioningState property is set to readOnly.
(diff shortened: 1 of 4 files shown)


docs/provisioning-state-must-be-read-only.md
```diff
@@ -8,16 +8,168 @@ ARM Error
...
+      },
+    },
+    "FooProps": {
+      "properties": {
+        "provisioningState": {
+        "$ref": "#/definitions/ProvisioningState",
+        "readOnly": true,
+        "type": "string",
+        "description": "Provisioning state of the foo rule.",
+        "enum": ["Creating", "Canceled", "Deleting", "Failed"],
```
rkmanda, line 94: pl remove these 3 lines for type, description and enum. This example should only have $ref and readOnly proeprties

rkmanda (approved):
:shipit:

docs/provisioning-state-must-be-read-only.md
AkhilaIlla: I thought, it could have the actual def. here

PR #643, opened 2023-12-19
Disable false-positive flagging rules in PROD

AkhilaIlla:
Folks have been complaining about rules being falsely flagged after the prod release.
This PR is to disable all those rules and also to fix RequestBodyMustExistForPutPatch(rules looks for the body param name to be "body" and it isn't mandatory for the body param's name to be "body", it is wrong and fixing it in this PR)


common/changes/@microsoft.azure/openapi-validator-rulesets/akhilailla-fix_linter_errors_2023-12-19-20-36.json
bdefoy, line 1: Add another changelog file for the RequestBodyMustExistForPutPatch changes. You can run rush change again and there should be an option to add another change

packages/rulesets/src/spectral/functions/utils.ts
```diff
@@ -90,7 +90,7 @@ export function getProperty(schema: any, propName: string): any {
 }
 
 export function findBodyParam(params: any) {
-  const isBody = (elem: any) => elem.name === "body" && elem.in === "body"
+  const isBody = (elem: any) => elem.name === "body"
```
tejaswiMinnu, line 93: What is this change for? is it auto generated?

packages/rulesets/src/spectral/test/request-body-must-exist-for-put-patch.test.ts
tejaswiMinnu, line 1: Any reason why to change "body" to "testBody" ?

packages/rulesets/src/spectral/functions/utils.ts
bdefoy, line 93: Are we sure we don't need `"in": "body"` for body parameters? It looks like in OpenAPI 2.0 `in` is [required for all parameters](https://swagger.io/specification/v2/#parameter-object), but in OpenAPI 3.0 body parameters use a different property name: `requestBody`.

packages/rulesets/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@microsoft.azure/openapi-validator-rulesets",
-  "version": "1.3.1",
+  "version": "1.3.2",
```
bdefoy, line 3: When updating the version make sure to use `rush version --bump` so the changelog gets generated. See https://github.com/Azure/azure-openapi-validator/blob/main/CONTRIBUTING.md#create-a-release-pr

packages/rulesets/src/spectral/functions/utils.ts
AkhilaIlla: Nope, it isnt mandatory for the body param's nam eto be "body", it is wrong and fixing it in this PR

packages/rulesets/src/spectral/test/request-body-must-exist-for-put-patch.test.ts
AkhilaIlla: just updating tests to check the fix I made

packages/rulesets/package.json
AkhilaIlla: good point !
Ill do that, thanks

packages/rulesets/src/spectral/functions/utils.ts
bdefoy: Oh, in that case shouldn't your change be:

```suggestion
  const isBody = (elem: any) => elem.in === "body"
```
since we don't care about the param name?

AkhilaIlla: My bad,
The intent was to remove the name===body check

AkhilaIlla: yup, just fixed it!
thanks

packages/rulesets/package.json
tejaswiMinnu: When you do version bump, you need to discard other file changes.

common/changes/@microsoft.azure/openapi-validator-core/FixDirectCycleIssue_2023-10-10-01-12.json
bdefoy, line 1: add this file back (i think you can use `git checkout`) since the openapi-validator-core package isn't being updated

common/changes/@microsoft.azure/openapi-validator-rulesets/akhilailla-Make_XmsIdentifierValidation_rule_warning_2023-12-04-20-10.json
bdefoy, line 1: add this file back (i think you can use `git checkout`) since the openapi-validator-core package isn't being updated

bdefoy (approved):

packages/rulesets/package.json
AkhilaIlla: @tejaswiMinnu  wdym by other file changes?

tejaswiMinnu: Few files were deleted .... [rkmanda-fixClientFlattenDocumentation_2023-12-09-02-48.json](https://github.com/Azure/azure-openapi-validator/pull/643/files#diff-812bc9dbff6f86bf7444a1b41ae539f7c6372527007e0912fcc35333b4a096e4)

AkhilaIlla: yeah, that is expected.
when we do the version bump, individual changeLog files get deleted and all those are consolidated and added to the main changelog.json

PR #644, opened 2024-01-11
Fix false positive for NoDuplicatePathsForScopeParameter

bdefoy:
`/{scope}/providers/Microsoft.Security/customRecommendations/{customRecommendationName}` and `/{scope}/providers/Microsoft.Security/customRecommendations` are identified as duplicated paths.

Added a `$` to the end of the regex to make the match more accurate.


rkmanda (approved):
:shipit:

PR #647, opened 2024-01-22
Tejaswis/fps properties type object no definition

tejaswiMinnu:
PropertiesTypeObjectNoDefinition must verify the existence of the "allOf" property in the definition object, previously the rule wasn't checking this condition which resulted to false positives.

WI - https://msazure.visualstudio.com/One/_workitems/edit/26124716

### DIFF
--- a/packages/rulesets/src/spectral/functions/properties-type-object-no-definition.ts
+++ b/packages/rulesets/src/spectral/functions/properties-type-object-no-definition.ts
@@ -19,7 +19,9 @@ export const propertiesTypeObjectNoDefinition: any = (definitionObject: any, opt
   }
 
   if (typeof definitionObject === "object") {
-    if (definitionObject.allOf) return
+    if (definitionObject.allOf) {
+      return
+    }
     if (definitionObject.properties === undefined) {
       if (definitionObject.additionalProperties === undefined) {
         return [{ message: errorMessageObject, path }]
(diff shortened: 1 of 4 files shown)


packages/rulesets/src/spectral/test/properties-type-object-no-definition.test.ts
```diff
@@ -220,12 +220,11 @@ test("PropertiesTypeObjectNoDefinition should find no errors", () => {
...
           "Common error response for all Azure Resource Manager APIs to return error details for failed operations. (This also follows the OData error response format.).",
         type: "object",
-        properties: {
-          error: {
-            description: "The error object.",
-            $ref: "#/definitions/ErrorDetail",
+        allOf: [
+          {
+            $ref: "#/definitions/ErrorAdditionalInfo",
           },
```
rkmanda, line 226: pl add a separate test for this variation

rkmanda (approved):
:shipit:

packages/rulesets/src/spectral/functions/properties-type-object-no-definition.ts
```diff
@@ -19,6 +19,7 @@ export const propertiesTypeObjectNoDefinition: any = (definitionObject: any, opt
   }
 
   if (typeof definitionObject === "object") {
+    if (definitionObject.allOf) return
```
rkmanda, line 22: >if (definitionObject.allOf) return [](http://example.com/codeflow?start=4&length=34)

format as 

    if (definitionObject.allOf) {
      return
     }

tejaswiMinnu: Formatted

PR #649, opened 2024-01-24
Fix ParameterNotUsingCommonTypes rule

tejaswiMinnu:
The existed ParameterNotUsingCommonTypes rule concatenates all the calls on a path instead of verifying individual GET/POST/PUT/DELETE call

Modified the rules to check each call separately and added a condition which checks all the parameters defined under $.paths and $.parameters 

https://msazure.visualstudio.com/One/_workitems/edit/26378654

### DIFF
--- a/common/changes/@microsoft.azure/openapi-validator-rulesets/tejaswis-fps-ParameterNotUsingCommonTypes_2024-01-24-18-56.json
+++ b/common/changes/@microsoft.azure/openapi-validator-rulesets/tejaswis-fps-ParameterNotUsingCommonTypes_2024-01-24-18-56.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@microsoft.azure/openapi-validator-rulesets",
+      "comment": "Fix an issue where ParameterNotUsingCommonTypes rule concatenates all the calls on a path instead of verifying individual GET/POST/PUT/DELETE call, these changes modify the rule accordingly",
+      "type": "patch"
+    }
+  ],
+  "packageName": "@microsoft.azure/openapi-validator-rulesets"
+}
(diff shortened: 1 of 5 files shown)


common/changes/@microsoft.azure/openapi-validator-rulesets/tejaswis-fps-ParameterNotUsingCommonTypes_2024-01-24-18-56.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@microsoft.azure/openapi-validator-rulesets",
+      "comment": "Fix an issue where ParameterNotUsingCommonTypes rule concatenates all the calls on a path instead of verifying individual GET/POST/PUT/DELETE call, these changes modify the rule accordingly",
```
bdefoy, line 5: I think this changelog can use some more clarity for end user perspective. What does the fix do directly for the end users? Was the 'path' output wrong previously, and now it correctly points to the right location?

bdefoy (approved):

common/changes/@microsoft.azure/openapi-validator-rulesets/tejaswis-fps-ParameterNotUsingCommonTypes_2024-01-24-18-56.json
tejaswiMinnu: Its not about pointing the location. The rule was previously concatenating all the parameters from all calls under a path and filtering was messed up. Though PUT doesn't have the for eg "api-version" and get has it, this rule is throwing error on PUT call because of concatenation. This is resulting the false positive.

rkmanda: You can describe the issue from a user's standpoint. Just mention the symptom of the problem instead of describing the problem in the code. You can say - "Fix a false alarm with the ParameterNotUsingCommonTypes rule that resulted in errors being flagged even when common-types were being referenced.

rkmanda (approved):
:shipit:

common/changes/@microsoft.azure/openapi-validator-rulesets/tejaswis-fps-ParameterNotUsingCommonTypes_2024-01-24-18-56.json
tejaswiMinnu: Addressed, thanks for the input

PR #650, opened 2024-01-26
Include RPCGuidelineCode to the linter output

konrad-jamrozik:


konrad-jamrozik:
@AkhilaIlla this PR fixes the issue you described [here](https://teams.microsoft.com/l/message/19:4f661242c446452e895359cc3ef45125@thread.tacv2/1706306079401?tenantId=72f988bf-86f1-41af-91ab-2d7cd011db47&groupId=0cab4ce9-7691-42ae-82e3-460d4346a710&parentMessageId=1706119062396&teamName=ARM%20API%20Reviewers&channelName=ARM%20linter%20rules&createdTime=1706306079401) plus does few other refactorings.

You mentioned you confirmed that `mergedResults` had the RPC guideline code here:
https://github.com/Azure/azure-openapi-validator/blob/580952ff6c8e9d60e9b3f0418abaf8a05ad79c37/packages/azure-openapi-validator/autorest/src/spectral-plugin-func.ts#L207

```
  const results = await spectral.run(doc)
  mergedResults.push(...results.map((result: any) => format(result, createFileOrFolderUri(filePath))))
  for (const message of mergedResults) {
    sendMessage(convertLintMsgToAutoRestMsg(message))
  }
```

The problem was inside `convertLintMsgToAutoRestMsg`, so 3 lines later. Overall, I do recommend to apply the technique of bisecting the problem: you knew that `mergedResults` had the value, and you knew that AutoRest is not printing it. Given that `sendMessage` is sending the results to AutoRest, this means the problem is likely somewhere between `mergedResults` and `sendMessage`. And indeed, it was there. Good technique to remember, in general :)
  return mergedResults

packages/azure-openapi-validator/autorest/src/spectral-plugin-func.ts
```diff
@@ -225,7 +213,9 @@ async function runSpectral(
   return mergedResults
 }
 
-export async function getRuleSet(openapiType: OpenApiTypes): Promise<[ruleset: Ruleset, ruleSetIncludingRPCGuidelineCode: any, namesOfRulesInStagingOnly: string[]]> {
+export async function getRuleSet(
+  openapiType: OpenApiTypes
+): Promise<[ruleset: Ruleset, c: any, namesOfRulesInStagingOnly: string[]]> {
```
AkhilaIlla, line 218: >c [](http://example.com/codeflow?start=30&length=1)

Can we have a descriptive variable name

AkhilaIlla (approved):

packages/azure-openapi-validator/autorest/src/spectral-plugin-func.ts
konrad-jamrozik: Oh wow sorry for that, that was not intended :D

AkhilaIlla (approved):

PR #663, opened 2024-02-24
Add support for disabling select Spectral rules for TypeSpec-generated specs

konrad-jamrozik:
This PR addresses:

- https://github.com/Azure/azure-sdk-tools/issues/7732

and subsumes:

- https://github.com/Azure/azure-openapi-validator/pull/622

Corresponding `openapi-alps` PR:
- [Pull Request 531407](https://devdiv.visualstudio.com/DevDiv/_git/openapi-alps/pullrequest/531407): LintDiff 2.2.0: support for disabling some Spectral rules for TypeSpec-generated specs

This PR:

- adds support to disable select Spectral rules for TypeSpec via a new property in given rule: `disableForTypeSpec: true`
- adds support for providing reason for that by another property: `disableForTypeSpecReason: "reason_here"`
- disables a few rules for TypeSpec as proposed by https://github.com/Azure/azure-openapi-validator/pull/622
- does a major refactoring and deduplication of code internals

### Testing

I ran LintDiff locally on my local clone of:

- https://github.com/Azure/azure-rest-api-specs/tree/main/specification/storage/resource-manager/Microsoft.Storage/stable/2023-01-01/storage.json

with following command line:

``` powershell
$use = "<path to my local packages/azure-openapi-validator/autorest>"
$apiVersion = "package-2023-01"
$autorestConfigFile = "<my local specs clone>/specification/storage/resource-manager/readme.md"
autorest --v3 --spectral --azure-validator --use=$use --tag=$apiVersion $autorestConfigFile
```

I also did variations of the specs file and the rulesets, including testing that:

- If the input spec file has `$.info.x-typespec-generated` then it will be recognized as TypeSpec-generated;
- Regression test: if a rule is staging only then it won't be included in prod run.
  - Also tested this for rules in common ruleset, applicable due to being extended by ARM ruleset.
- If a rule is disabled for TypeSpec, it is in fact disable if and only if the spec is considered to be TypeSpec-generated.
  - Also tested for rules extended from common ruleset.


packages/rulesets/src/spectral/az-arm.ts
```diff
@@ -988,6 +988,8 @@ const ruleset: any = {
       description: "This rule checks for references that aren't using latest version of common-types.",
       message: "{{error}}",
       severity: "warn",
+      disableForTypeSpec: true,
+      disableForTypeSpecReason: "TODO",
```
konrad-jamrozik, line 992: @tjprescott @johanste @timotheeguerin if you folks could help me provide values for all `disableForTypeSpecReason: TODO`, that would be awesome :)

weshaggard:
It should be a goal to try an remove these suppressions in our sample project once these rules are disabled https://github.com/Azure/azure-rest-api-specs/blob/main/specification/contosowidgetmanager/data-plane/readme.md#suppress-non-typespec-sdk-related-linting-rules

packages/rulesets/generated/spectral/az-arm.js
```diff
@@ -844,6 +844,8 @@ const ruleset$1 = {
             description: "Operation should have a summary or description.",
             message: "Operation should have a summary or description.",
             severity: "warn",
+            disableForTypeSpec: true,
```
weshaggard, line 847: While I'm fine with hardcoding these configurations for now I do think we need to work towards a goal of having these configurations be external of the tool so we don't need to release a new tool to enable or disable specific rules.

packages/azure-openapi-validator/autorest/src/spectral-plugin-func.ts
```diff
@@ -90,189 +65,114 @@ export async function spectralPluginFunc(initiator: IAutoRestPluginInitiator): P
...
-    .sort(Intl.Collator().compare)
+  try {
+    const openApiSpecFilePath = openApiSpecFile.startsWith("file:///") ? fileURLToPath(openApiSpecFile) : openApiSpecFile
+    const openApiSpecContent: string = await readFileUsingCache(initiator, openApiSpecFile)
+    // safeLoad() documented at: https://github.com/nodeca/js-yaml/tree/v3?tab=readme-ov-file#safeload-string---options-
+    // Empirically confirmed the returned value type is object, not string.
+    const openApiSpecYml: any = safeLoad(openApiSpecContent)
+    // "x-typespec-generated" is expected to be found at JSONPath of $.info.x-typespec-generated.
+    // Example: https://github.com/Azure/azure-rest-api-specs/blob/fca48bec19cc5aab0a45c0769bfca0f667164dbf/specification/edgemarketplace/resource-manager/Microsoft.EdgeMarketplace/stable/2023-08-01/operations.json#L7
+    const specIsGeneratedFromTypeSpec = Boolean(openApiSpecYml.info["x-typespec-generated"]) // JSON.stringify(openApiSpecYml).includes("x-typespec-generated")
```
weshaggard, line 129: Do we need to handle the case where before was not typespec but the after is typespec now?

packages/rulesets/generated/spectral/az-arm.js
```diff
@@ -862,6 +864,8 @@ const ruleset$1 = {
             description: "All schemas should have a description or title.",
             message: "Schema should have a description or title.",
             severity: "warn",
+            disableForTypeSpec: true,
+            disableForTypeSpecReason: "Covered by TSP's 'documentation-required' rule.",
```
timotheeguerin, line 868:
```suggestion
            disableForTypeSpecReason: "Covered by TSP's '@azure-tools/typespec-azure-core/documentation-required' rule.",
```

not sure how much this matter but if you want to be accurate in the reference this is the full name of the rule

packages/rulesets/generated/spectral/az-arm.js
```diff
@@ -1186,6 +1192,8 @@ const ruleset$1 = {
             description: "The value of the 'description' property must be descriptive. It cannot be spaces or empty description.",
             message: "'{{property}}' parameter lacks 'description' property. Consider adding a 'description' element. Accurate description is essential for maintaining reference documentation.",
             severity: "error",
+            disableForTypeSpec: true,
+            disableForTypeSpecReason: "TODO",
```
timotheeguerin, line 1196:
```suggestion
            disableForTypeSpecReason: "Covered by TSP's '@azure-tools/typespec-azure-core/documentation-required' rule.",
```

I believe this is the same rule

packages/rulesets/generated/spectral/az-arm.js
```diff
@@ -3710,6 +3718,8 @@ const ruleset = {
             description: "This rule checks for references that aren't using latest version of common-types.",
             message: "{{error}}",
             severity: "warn",
+            disableForTypeSpec: true,
+            disableForTypeSpecReason: "TODO",
```
timotheeguerin, line 3722: This one I am not fully sure we have a linter rule for that does exactly that. We have one that ensure you picked a version but not check we use the latest.

packages/rulesets/generated/spectral/az-common.js
```diff
@@ -613,6 +613,8 @@ const ruleset = {
             description: "Operation should have a summary or description.",
             message: "Operation should have a summary or description.",
             severity: "warn",
+            disableForTypeSpec: true,
+            disableForTypeSpecReason: "TODO",
```
timotheeguerin, line 617:
```suggestion
            disableForTypeSpecReason: "Covered by TSP's '@azure-tools/typespec-azure-core/documentation-required' rule.",
```

I believe this is the same rule

packages/rulesets/generated/spectral/az-common.js
```diff
@@ -955,6 +961,8 @@ const ruleset = {
             description: "The value of the 'description' property must be descriptive. It cannot be spaces or empty description.",
             message: "'{{property}}' parameter lacks 'description' property. Consider adding a 'description' element. Accurate description is essential for maintaining reference documentation.",
             severity: "error",
+            disableForTypeSpec: true,
+            disableForTypeSpecReason: "TODO",
```
timotheeguerin, line 965:
```suggestion
            disableForTypeSpecReason: "Covered by TSP's '@azure-tools/typespec-azure-core/documentation-required' rule.",
```

I believe this is the same rule

packages/rulesets/generated/spectral/az-dataplane.js
```diff
@@ -643,6 +643,8 @@ const ruleset$1 = {
             description: "Operation should have a summary or description.",
             message: "Operation should have a summary or description.",
             severity: "warn",
+            disableForTypeSpec: true,
+            disableForTypeSpecReason: "TODO",
```
timotheeguerin, line 647:
```suggestion
            disableForTypeSpecReason: "Covered by TSP's '@azure-tools/typespec-azure-core/documentation-required' rule.",
```

I believe this is the same rule

packages/rulesets/generated/spectral/az-dataplane.js
```diff
@@ -985,6 +991,8 @@ const ruleset$1 = {
             description: "The value of the 'description' property must be descriptive. It cannot be spaces or empty description.",
             message: "'{{property}}' parameter lacks 'description' property. Consider adding a 'description' element. Accurate description is essential for maintaining reference documentation.",
             severity: "error",
+            disableForTypeSpec: true,
+            disableForTypeSpecReason: "TODO",
```
timotheeguerin, line 995:
```suggestion
            disableForTypeSpecReason: "Covered by TSP's '@azure-tools/typespec-azure-core/documentation-required' rule.",
```

I believe this is the same rule

packages/rulesets/generated/spectral/az-dataplane.js
```diff
@@ -1942,6 +1950,8 @@ const ruleset = {
     rules: {
         AdditionalPropertiesAndProperties: {
             description: "Don't specify additionalProperties as a sibling of properties.",
+            disableForTypeSpec: true,
+            disableForTypeSpecReason: "TODO",
```
timotheeguerin, line 1954:
```suggestion
            disableForTypeSpecReason: "Covered by TSP's '@azure-tools/typespec-azure-core/bad-record-type' rule.",
```

I believe this is the same rule

packages/rulesets/generated/spectral/az-dataplane.js
```diff
@@ -2064,6 +2074,8 @@ const ruleset = {
         Nullable: {
             description: "Avoid the use of x-nullable.",
             severity: "warn",
+            disableForTypeSpec: true,
+            disableForTypeSpecReason: "TODO",
```
timotheeguerin, line 2078:
```suggestion
            disableForTypeSpecReason: "Covered by Azure Core '@azure-tools/typespec-azure-core/no-nullable' rule",
```

packages/rulesets/generated/spectral/az-dataplane.js
```diff
@@ -2190,6 +2202,8 @@ const ruleset = {
             description: "All schema properties should have a description.",
             message: "Property should have a description.",
             severity: "warn",
+            disableForTypeSpec: true,
+            disableForTypeSpecReason: "TODO",
```
timotheeguerin, line 2206:
```suggestion
            disableForTypeSpecReason: "Covered by TSP's '@azure-tools/typespec-azure-core/bad-record-type' rule.",
```

I believe this is the same rule

packages/rulesets/generated/spectral/az-dataplane.js
```diff
@@ -2201,6 +2215,8 @@ const ruleset = {
             description: "All schema properties should have a defined type.",
             message: "Property should have a defined type.",
             severity: "warn",
+            disableForTypeSpec: true,
+            disableForTypeSpecReason: "TODO",
```
timotheeguerin, line 2219: Not exactly sure what this rule check but this might just be covered by typespec syntax that requires you to give a type to a property.

packages/rulesets/src/spectral/az-arm.ts
timotheeguerin: I see there is a lot of duplicate rules in different ruleset, are they exactly the same?

tjprescott: I can do that. Thank you @konrad-jamrozik!

packages/rulesets/generated/spectral/az-arm.js
timotheeguerin: Wondering if that would not be beneficial to do right now as it seems rulesets are all redefining the same disableForTypeSpec and reason, maybe it would be better to hard code a map for now so we can write the reason once.

packages/rulesets/generated/spectral/az-arm.js
tjprescott: Also the arm-common-types rule is not currently enabled by default, so I don't think we should disable this yet.

packages/rulesets/generated/spectral/az-dataplane.js
tjprescott: Actually we don't have this validation in place on the TSP side:
https://github.com/Azure/typespec-azure-pr/issues/3882

I don't think we should disable this rule yet.

packages/rulesets/generated/spectral/az-dataplane.js
tjprescott: No this is the documentation-required rule.

packages/rulesets/generated/spectral/az-dataplane.js
timotheeguerin: Oh I thought we had that rule but yeah apparently not. So yeah lets' not remove yet.

packages/rulesets/generated/spectral/az-dataplane.js
timotheeguerin: hhm copy paste issue probably

packages/rulesets/generated/spectral/az-common.js
```diff
@@ -631,6 +633,8 @@ const ruleset = {
             description: "All schemas should have a description or title.",
             message: "Schema should have a description or title.",
             severity: "warn",
+            disableForTypeSpec: true,
+            disableForTypeSpecReason: "Covered by TSP's 'documentation-required' rule.",
```
tjprescott, line 637:
```suggestion
            disableForTypeSpecReason: "Covered by TSP's '@azure-tools/typespec-azure-core/documentation-required' rule.",
```

packages/rulesets/generated/spectral/az-common.js
```diff
@@ -641,6 +645,8 @@ const ruleset = {
             description: "All parameters should have a description.",
             message: "Parameter should have a description.",
             severity: "warn",
+            disableForTypeSpec: true,
+            disableForTypeSpecReason: "Covered by TSP's 'documentation-required' rule.",
```
tjprescott, line 649:
```suggestion
            disableForTypeSpecReason: "Covered by TSP's '@azure-tools/typespec-azure-core/documentation-required' rule.",
```

packages/rulesets/generated/spectral/az-common.js
tjprescott: I'm curious how "ParameterDescriptionRequired" is not duplicative of "ParameterDescription"?  Either way, they can both be disabled.

packages/rulesets/generated/spectral/az-dataplane.js
```diff
@@ -661,6 +663,8 @@ const ruleset$1 = {
             description: "All schemas should have a description or title.",
             message: "Schema should have a description or title.",
             severity: "warn",
+            disableForTypeSpec: true,
+            disableForTypeSpecReason: "Covered by TSP's 'documentation-required' rule.",
```
tjprescott, line 667:
```suggestion
            disableForTypeSpecReason: "Covered by TSP's '@azure-tools/typespec-azure-core/documentation-required' rule.",
```

packages/rulesets/generated/spectral/az-dataplane.js
```diff
@@ -671,6 +675,8 @@ const ruleset$1 = {
             description: "All parameters should have a description.",
             message: "Parameter should have a description.",
             severity: "warn",
+            disableForTypeSpec: true,
+            disableForTypeSpecReason: "Covered by TSP's 'documentation-required' rule.",
```
tjprescott, line 679:
```suggestion
            disableForTypeSpecReason: "Covered by TSP's '@azure-tools/typespec-azure-core/documentation-required' rule.",
```

packages/azure-openapi-validator/autorest/src/spectral-plugin-func.ts
weshaggard: I'm not sure what we want to do here but perhaps we can check to see if the after is typespec and disable the rules for both the before and after runs.

tjprescott (commented):
Looks wonderful to me! I left some comments on some of the rules and some I don't think we should disable yet.

packages/rulesets/generated/spectral/az-arm.js
konrad-jamrozik: @weshaggard @timotheeguerin out of scope for now - I need to deliver this by Thursday. I am happy to work on any improvements in a follow-up PR.

packages/azure-openapi-validator/autorest/src/spectral-plugin-func.ts
konrad-jamrozik: @weshaggard from the point of view of LintDiff I believe we won't ever produce any extra errors. If a spec was converted to TypeSpec it can only have **less** errors reported, because we are only **disabling** rules for TypeSpec case. If I understand @tjprescott correctly, there might be a case the new spec gets new errors, but not from LintDiff, but the built-in TypeSpec linter. This will be especially noticeable if there was a LintDiff violation suppressed in the "before/manual" spec, but the "after/TypeSpec-generated" spec will be flagged on the same problem  by the TypeSpec linter, because the TypeSpec linter does not read LintDiff suppressions. In such cases we will have to add TSP suppressions, the way @tjprescott explained.

weshaggard: Ok the less errors case makes sense and covers my concerns.

packages/rulesets/src/spectral/az-arm.ts
konrad-jamrozik, line 992: @tjprescott @timotheeguerin  what should be the explanation for disabling `LatestVersionOfCommonTypesMustBeUsed` ?

konrad-jamrozik: @timotheeguerin also - you folks were commenting on `.js` files instead of `.ts` files. `.js` files with rules are generated from `.ts` files by merging `common` ruleset with `arm` or `dataplan` ruleset. In the resulting `.js` files some rules are duplicated.

tjprescott: We should not disable it.

packages/rulesets/generated/spectral/az-arm.js
tjprescott: No, we should not disable in this PR.

weshaggard (approved):

PR #686, opened 2024-04-26
[ValidFormats] Only validate "format" when "type" is expected primitive type

### LINKED ISSUE #685: [TypeSpec False Positive] TypeSpec model contains property named "format"
TypeSpec spec has a model propery named "format":

```
model Glossary {
  @doc("Format")
  format: string;
```

https://github.com/Azure/azure-rest-api-specs/pull/28239/files#diff-d8b43d27a4b5a6962c6021795ff2c550365077c6170b7e12ce742915644533a3R202

Which causes this LintDiff error:

image

https://github.com/Azure/azure-openapi-validator/blob/main/docs/valid-formats.md

Which we believe is a false positive.  The rule should apply to the `format` property within a schema, but not to a model property that happens to be named `format`.

We would need to dismbiguate these cases, which may be difficult (or even impossible) given the current design.  If too difficult, resolve this as "won't fix" and recommend suppressions for this edge case.

## Ignore "format"
```
"FileFormat": {
  "type": "object",
  "description": "File Format",
  "properties": {
    "format": {
      "type": "string",
      "description": "Name of the format"
    },
```

https://github.com/Azure/azure-rest-api-specs/blob/e84efe1a891af21216562b4fac72b18e5662c78c/specification/translation/data-plane/Azure.AI.DocumentTranslation/stable/2024-05-01/openapi.json#L877-L884

## Validate "format"
```
{
  "name": "top",
  "in": "query",
  "description": "...",
  "required": false,
  "type": "integer",
  "format": "int32",
  "maximum": 2147483647
},
```

https://github.com/Azure/azure-rest-api-specs/blob/e84efe1a891af21216562b4fac72b18e5662c78c/specification/translation/data-plane/Azure.AI.DocumentTranslation/stable/2024-05-01/openapi.json#L72-L80

## Possible Fixes

1. Change `given` to select only properties we want to validate:

https://github.com/Azure/azure-openapi-validator/blob/fb507f20784f72ddeeeda1486aeef8f64829c8bd/packages/rulesets/src/spectral/az-common.ts#L231

2. Change function `checkSchemaFormat` to ignore properties we **don't** want to validate.

https://github.com/Azure/azure-openapi-validator/blob/fb507f20784f72ddeeeda1486aeef8f64829c8bd/packages/rulesets/src/spectral/functions/schema-format.ts#L10

mikeharder:
- Fixes #685

Per https://github.com/OAI/OpenAPI-Specification/blob/main/versions/2.0.md#data-types, I believe we should only validate properties named `format` if there is a sibling property named `type` with value `"boolean" | "integer" | "number" | "string"`.


packages/rulesets/src/spectral/az-common.ts
```diff
@@ -228,7 +228,7 @@ const ruleset: any = {
       message: "'{{error}}' is not a known format.",
       severity: "error",
       resolved: false,
-      given: "$..[?(@property === 'format')]^",
```
mikeharder, line 231: Previous query:
* Find all properties named `format`
* Return their parent object (`^`)

New query:
* Find all objects with a property named named `format` AND
* A property named `type` with value `"boolean" | "integer" | "number" | "string"`
* Don't need to return parent using `^` since we are searching objects, not properties

konrad-jamrozik:
Don't forget to bump the version of the ruleset in the package.json

konrad-jamrozik (approved):

konrad-jamrozik (approved):

PR #694, opened 2024-05-28
Fix false alarm for EvenSegmentedPathForPutOperation rule

AkhilaIlla:
The rule checks for even segments in the path and in addition checks if the path ends with {} & 
flagged paths like "/providers/Microsoft.PortalServices/copilotSettings/default" which is still valid.

This PR addresses such issues by considering default in addition to the fields ending with {}

### DIFF
--- a/packages/rulesets/src/spectral/az-arm.ts
+++ b/packages/rulesets/src/spectral/az-arm.ts
@@ -601,7 +601,8 @@ const ruleset: any = {
       then: {
         function: pattern,
         functionOptions: {
-          match: ".*/providers/\\w+.\\w+(/\\w+/{\\w+})+$",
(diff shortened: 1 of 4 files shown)


packages/rulesets/src/spectral/az-arm.ts
```diff
@@ -601,7 +601,8 @@ const ruleset: any = {
       then: {
         function: pattern,
         functionOptions: {
-          match: ".*/providers/\\w+.\\w+(/\\w+/{\\w+})+$",
+          // match: ".*/providers/[\\w\\.]+(?:/\\w+/(default|{\\w+}))*/\\w+$",
+          match: ".*/providers/\\w+.\\w+(/\\w+/(default|{\\w+}))+$",
         },
```
rkmanda, line 606: expand the logic to terms other than default in a separate PR

rkmanda (approved):
:shipit:

PR #695, opened 2024-05-28
Fix false alarm for XmsPageableForListCalls rule

AkhilaIlla:
The rule checks if the list calls have XmsPageableForListCalls  field & hence the logic tries to get all the paths that dont end with "}" &
flags paths like "/providers/Microsoft.PortalServices/copilotSettings/default" which is still valid.

This PR addresses such issues.

### DIFF
--- a/packages/rulesets/generated/spectral/az-arm.js
+++ b/packages/rulesets/generated/spectral/az-arm.js
@@ -3250,7 +3250,7 @@ const ruleset = {
             message: "{{error}}",
             resolved: true,
             formats: [oas2],
-            given: "$[paths,'x-ms-paths'][?(!@property.endsWith('}'))].get",
+            given: "$[paths,'x-ms-paths'][?(!@property.endsWith('}') && !@property.endsWith('default'))].get",
             then: {
                 function: xmsPageableForListCalls,
             },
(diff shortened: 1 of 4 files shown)


packages/rulesets/generated/spectral/az-arm.js
```diff
@@ -3250,7 +3250,7 @@ const ruleset = {
             message: "{{error}}",
             resolved: true,
             formats: [oas2],
-            given: "$[paths,'x-ms-paths'][?(!@property.endsWith('}'))].get",
+            given: "$[paths,'x-ms-paths'][?(!@property.endsWith('}') && !@property.endsWith('/default'))].get",
```
rkmanda, line 3253: >default [](http://example.com/codeflow?start=94&length=7)

There could be other terms that maybe used - not just default. You can get this in for now as the maority case is when the term 'default' is sed. Pl open another item to cover for other cases. A better approach could be to make this check for odd segmented GET paths

packages/rulesets/src/spectral/az-arm.ts
```diff
@@ -418,7 +418,7 @@ const ruleset: any = {
       message: "{{error}}",
       resolved: true,
       formats: [oas2],
-      given: "$[paths,'x-ms-paths'][?(!@property.endsWith('}'))].get",
+      given: "$[paths,'x-ms-paths'][?(!@property.endsWith('}') && !@property.endsWith('/default'))].get",
       then: {
```
rkmanda, line 422: There could be other terms that maybe used - not just default. You can get this in for now as the maority case is when the term 'default' is sed. Pl open another item to cover for other cases. A better approach could be to make this check for odd segmented GET paths

rkmanda (approved):
:shipit:

PR #669, opened 2024-03-05
Disable LintDiff rules for TypeSpec 0.55

tjprescott:
Disables the following rules:
- PutResponseCodes
- PostResponseCodes
- ParameterDescriptionRequired
- LroLocationHeader
- ResrouceNameRestriction

PR #709, opened 2024-06-13
Fix BodyTopLevelPropertiesOnly to support get-only APIs

bdefoy:
BodyTopLevelProperties currently does not catch invalid top-level properties for get only APIs. This PR adds behavior to the rule to examine get operations in addition to the put and patch operations that it already examined.


packages/rulesets/generated/spectral/az-dataplane.js
bdefoy, line 1: Note for reviewers: these changes were generated when I ran the build.

rkmanda (approved):
:shipit:

packages/rulesets/generated/spectral/az-dataplane.js
mikeharder: I think this is fallout from https://github.com/Azure/azure-openapi-validator/pull/703.

@bdefoy: Please revert the changes to az-dataplane.js manually in your PR.

@konrad-jamrozik: We will need a better solution for the long term.

mikeharder (changes requested):
Revert changes to az-dataplane.js

mikeharder (changes requested):
I think you also want to bump the package version here, so you can publish a new version after merging:

https://github.com/azure/azure-openapi-validator/blob/main/packages/rulesets/package.json#L3

konrad-jamrozik (approved):

konrad-jamrozik:
@bdefoy note the current deployment process is here:

https://github.com/Azure/azure-openapi-validator/pull/669/files#r1515048617

mikeharder (approved):

PR #718, opened 2024-06-24
Akhilailla/fix properties type object no definition  false negatives

AkhilaIlla:
Blocking comment related to property of type object not having definition is added manually: (https://github.com/Azure/azure-rest-api-specs/pull/28891#discussion_r1597228278)

This PR validates if the false negative mentioned in https://msazure.visualstudio.com/One/_workitems/edit/28199177 (https://github.com/Azure/azure-rest-api-specs/pull/28891#discussion_r1597228278) is being flagged.

Also, found that no jobs ran until a couple of comments on the PR. Either the checks didnt run & hence didnt flag the error or the builds arent showing up now.
image


packages/rulesets/src/spectral/functions/properties-type-object-no-definition.ts
```diff
@@ -1,23 +1,23 @@
...
-const errorMessageNull =
-  "Properties with 'type' NULL are not allowed, please specify the 'type' as 'Primitive' or 'Object' referring a model."
+
+const errorMessageObject = "Properties with type:object that don't reference a model definition are not allowed. ARM doesn't allow generic type definitions as this leads to bad customer experience."
+const errorMessageNull = "Properties with 'type' NULL are not allowed, please specify the 'type' as 'Primitive' or 'Object' referring a model."
+
 export const propertiesTypeObjectNoDefinition: any = (definitionObject: any, opts: any, ctx: any) => {
   const path = ctx.path || []
 
+  // return errorMessageNull if type or properties are null
```
bdefoy, line 10: Is putting the check on `definitionObject?.type` before the check on `definitionObject?.properties` necessary to resolve the false negative in your test? I'm not sure I'm following how that change resolved the false negative

AkhilaIlla: No, the reordering isn't to fix the error.

It was just to keep all the property related checks in a place.

Also, this PR adds a test that was flagged as false negative & the method seems to work fine.

tejaswiMinnu (approved):
The reason this issue wasn't detected is because the rule is currently only in the staging environment. I have a PBI in this sprint to update the rules in the production environment.

PR #728, opened 2024-08-01
Added a check if 202 response codes have response schema defined when it shouldnt

tejaswiMinnu:
Added an if statement to make sure there should be no "schema" defined for 202 response codes for both PATCH and DELETE.
POST is already implemented, added similar if condition for both PATCH and DELETE

WI - https://msazure.visualstudio.com/One/_workitems/edit/28118100

### DIFF
--- a/packages/rulesets/src/spectral/functions/patch-response-codes.ts
+++ b/packages/rulesets/src/spectral/functions/patch-response-codes.ts
@@ -9,7 +9,7 @@ const LR_ERROR =
   "Long-running PATCH operations must have responses with 200, 202 and default return codes. They also must not have other response codes."
 const EmptyResponse_ERROR =
   "PATCH operation response codes must be non-empty. It must have response codes 200 and default if it is sync or 200, 202 and default if it is long running."
-  
+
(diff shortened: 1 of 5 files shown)


packages/rulesets/src/spectral/functions/patch-response-codes.ts
```diff
@@ -44,6 +46,13 @@ export const patchResponseCodes = (patchOp: any, _opts: any, ctx: any) => {
         path: path,
       })
     }
+
+    if (patchOp.responses["202"]?.schema) {
+      errors.push({
+        message: "202 response for a LRO PATCH operation must not have a response schema specified.",
+        path: path,
```
rkmanda, line 53: pl update the documentation as well to call out this condition.

rkmanda (changes requested):
:clock1:

packages/rulesets/src/spectral/functions/patch-response-codes.ts
tejaswiMinnu: Updated

rkmanda (approved):
:shipit:

Issue #707, opened 2024-06-11
Update the version format check for canonical swagger


dinwa-ms:
### Lint rule description

For canonical swagger, the version info is "canonical", which violates the existing rule of the formate "Year-Month-Date". Can we make this a special case for canonical swagger?

![image](https://github.com/Azure/azure-openapi-validator/assets/8683976/cd167336-5f9c-4800-880a-126c2780a3e0)

### Related swagger example

{
  "swagger": "2.0",
  "info": {
    "title": "Microsoft.Contoso management service",
    "version": "canonical",
    "description": "Microsoft.Contoso Resource Provider management API.",
    "x-canonical-included-versions": [
      "v2021_10_01_preview"
    ],
    "x-typespec-generated": [
      {
        "emitter": "@azure-tools/typespec-autorest-canonical"
      }
    ]
  },
  "schemes": [
    "https"
  ],
...

### Category

SDK

### Severity level

Error

### Applies to

Management plane API spec, Data plane API spec, RPaaS API spec

### How to fix the violation

For canonical swagger, set the version to "canonical"; for versioned swagger, set the version to the specific version.

### What't the impact if breaking the rule

The information will be misleading.

dinwa-ms:
Could you help provide an update on this issue? Thanks

mikeharder:
@rkmanda @bdefoy @tejaswiMinnu @AkhilaIlla: I believe this is an ARM-only rule:

https://github.com/Azure/azure-openapi-validator/blob/f3fcf47c9f74b15e90e5198bf09bb6569e13213e/packages/rulesets/src/spectral/az-arm.ts#L1066

dinwa-ms:
Hi @rkmanda @bdefoy @tejaswiMinnu @AkhilaIlla, may I ask is anyone working on this issue and is there any update or ETA on it? Thanks.

dinwa-ms:
If it hasn't started, I can try to work on it and will share the code change for review.

PR #732, opened 2024-08-21
Update arm APIVersionPattern rule for canonical swagger

dinwa-ms:
It is for [Issue 707 ](https://github.com/Azure/azure-openapi-validator/issues/707): Update the version format check for canonical swagger #707.

We introduced canonical swagger. It is a union of versioned swaggers. The version of canonical swagger will be "canonical",  which will fail the existing APIVersionPattern. In this PR, we updated the rule for this scenario. If the swagger is generated by autorest-canonical emitter, it is canonical swagger and its version should be 'canonical'. Otherwise, its version should follow the existing rule.


dinwa-ms:
@rkmanda @mikeharder, could you help take a look at this PR when available? Thanks

PR #733, opened 2024-08-23
Removed stagedonly tag and updated the package version

tejaswiMinnu:
Removed stagedonly tag and updated the package version


packages/rulesets/generated/spectral/az-arm.js
```diff
@@ -3354,7 +3354,6 @@ const ruleset = {
             rpcGuidelineCode: "RPC-Policy-V1-03",
             description: "Properties with type:object that don't reference a model definition are not allowed. ARM doesn't allow generic type definitions as this leads to bad customer experience.",
```
markcowl, line 3355: I agree with this one, but we do not have a similar rule inside TypeSpec.  Filed this issue to resolve: https://github.com/Azure/typespec-azure/issues/1426

markcowl (approved):

packages/rulesets/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@microsoft.azure/openapi-validator-rulesets",
-  "version": "2.0.10",
+  "version": "2.0.11",
```
mikeharder, line 3:
```suggestion
  "version": "2.1.0",
```

Let's start following semver.  Increment the minor when adding new rules/features, and increment patch if the only change is fixing bugs in existing rules.

mikeharder (changes requested):

packages/rulesets/package.json
tejaswiMinnu: Sure will do, thanks

mikeharder (approved):

PR #745, opened 2024-09-12
Add QueryParametersInCollectionGet linter rule

AkhilaIlla:
Add QueryParametersInCollectionGet linter rule to validate that list call should only have api-version and OData filter as query params.

### DIFF
--- a/docs/query-parameters-in-collection-get.md
+++ b/docs/query-parameters-in-collection-get.md
@@ -1,0 +1,78 @@
+# QueryParametersInCollectionGet
+
+## Category
+
+ARM Error
+
+## Applies to
+
+ARM OpenAPI(swagger) specs
+
+## Related ARM Guideline Code
+
+- RPC-Get-V1-15
+
+## Description
+
+Collection Get's/List operations MUST not have query parameters other than api-version & OData filter.
+
+## How to fix the violation
+
+Ensure that collection GET/List operations do not include any query parameters other than api-version and the OData $filter.
+
+## Good Examples
+
+```json
+      "Microsoft.Music/Songs": {
+        get: {
+          operationId: "Foo_Update",
+          description: "Test Description",
+          parameters: [
+            {
+              name: "api-version",
+              in: "query"
+            },
+          ],
+        },
+      },
+```
+
+```json
+      "Microsoft.Music/Songs": {
+        get: {
+          operationId: "Foo_Update",
+          description: "Test Description",
+          parameters: [
+            {
+              name: "api-version",
+              in: "query"
+            },
+            {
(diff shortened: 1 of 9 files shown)


docs/query-parameters-in-collection-get.md
```diff
@@ -0,0 +1,78 @@
...
+
+## How to fix the violation
+
+Ensure that collection GET/List operations do not include any query parameters other than api-version and the OData $filter.
+
+## Good Examples
+
+```json
+      "Microsoft.Music/Songs": {
+        get: {
```
bdefoy, line 27: nit: fix JSON in this section for better readability. e.g. 
```suggestion
        "get": {
```

packages/rulesets/src/spectral/functions/query-parameters-in-collection-get.ts
```diff
@@ -0,0 +1,38 @@
...
+  const errors: any[] = []
+
+  for (const uri of uris) {
+    // skip, if GET op isn't defined
+    if (!pathItem[uri][GET]) {
+      continue
+    }
+    // check if the GET op is a collection get/list call
+    const hierarchy = getResourcesPathHierarchyBasedOnResourceType(uri)
+    if (hierarchy.length == 0 && pathItem[uri][GET]) {
```
bdefoy, line 23: I think this is not needed since it's checked on line 18 already
```suggestion
    if (hierarchy.length == 0) {
```

daki90 (approved):

packages/rulesets/src/spectral/functions/query-parameters-in-collection-get.ts
```diff
@@ -0,0 +1,33 @@
...
+  const uris = Object.keys(pathItem)
+  if (uris.length < 1) {
+    return []
+  }
+  const GET = "get"
+  const errors: any[] = []
+
+  for (const uri of uris) {
+    //check if GET op is defined & the GET op is a collection get/list call
+    if (pathItem[uri][GET] && isListOperation(uri)) {
```
bdefoy, line 18: Good that you are checking for `get` here but `isListOperation` should probably be checking if it's a `get` operation. Would it be possible to add the check there? Or if it's too much effort this is fine for now. I haven't checked how many times `isListOperation` is used throughout the code

bdefoy (approved):

packages/rulesets/src/spectral/functions/query-parameters-in-collection-get.ts
AkhilaIlla: Thats a good point.
Lemme update.

PR #748, opened 2024-10-11
Tejaswis/remove rpc async v1 07

tejaswiMinnu:
Removed rule RPC-Async-V1-07


rkmanda:
For other reviewers: This rule is being removed to be inline with ARM's change of expectations where we want to deprecate the use of the location header and promote the use of the Async operation header instead.

packages/rulesets/src/spectral/az-arm.ts
```diff
@@ -11,7 +11,6 @@ import { DeleteResponseCodes } from "./functions/delete-response-codes"
 import { getCollectionOnlyHasValueAndNextLink } from "./functions/get-collection-only-has-value-and-next-link"
 import { getResponseCodes } from "./functions/get-response-codes"
 import hasApiVersionParameter from "./functions/has-api-version-parameter"
-import hasheader from "./functions/has-header"
```
bdefoy, line 14: Also remove has-header.cs if it's no longer used

https://github.com/Azure/azure-openapi-validator/blob/main/packages/rulesets/src/spectral/functions/has-header.ts

bdefoy: never mind, i see it is still used

bdefoy (approved):

PR #762, opened 2025-01-21
Skip validating MSI in PatchBodyParametersSchema

AkhilaIlla:
This PR skips validation for MSI(managed service identity) as it is being referenced from common-types & has the required field & is being referenced in patch body parameter schema by several RP's & are having to get an exception. So, skipping to avoid that step.

Issue #750, opened 2024-10-16
[GetCollectionResponseSchema] Bugs detecting/reporting errors in APIs that vary only by subscription prefix


mikeharder:
Rule `GetCollectionResponseSchema` may have one or more bugs, related to detecting and/or reporting errors in APIs that vary only by the subscription prefix.  For more details, see this thread:

https://github.com/Azure/azure-rest-api-specs/pull/30656#discussion_r1802351631

mikeharder:
It turns out this behavior is in YAML itself, not our code.  YAML has types for strings and timestamps, and an unquoted string is parsed as a timestamp if possible.

```yaml
value-parsed-as-string-1: "2024-05-01"
value-parsed-as-string-2: foo2024-05-01
value-parsed-as-timestamp: 2024-05-01
```

mikeharder:
/subscriptions/{subscriptionId}/providers/Microsoft.HybridCompute/locations/{location}/publishers/{publisher}/extensionTypes/{extensionType}/versions/{version}
operationId: `ExtensionMetadata_Get`
Response: `#/definitions/ExtensionValue`

/subscriptions/{subscriptionId}/providers/Microsoft.HybridCompute/locations/{location}/publishers/{publisher}/extensionTypes/{extensionType}/versions
operationId: `ExtensionMetadata_List`
Response: `#/definitions/ExtensionValueListResult`

```
    "ExtensionValueListResult": {
      "type": "object",
      "properties": {
        "value": {
          "type": "array",
          "readOnly": true,
          "items": {
            "$ref": "#/definitions/ExtensionValue"
          },
```

bdefoy:
From https://github.com/Azure/azure-rest-api-specs/pull/30656#issuecomment-2427379648:

> Since the schema is different, the name of the Type must also be different within a namespace. The ARM Manifest would have a single entry for the type and doesnt document the scopes separately, so it makes even more sense to not allow the pattern where the definition varies at each scope.

bdefoy:
Since ARM considers the V1 and V2 endpoints the same resource type, the collection get and point get should have the same response schema across both V1 and V2. With that in mind, the rule is valid to flag here because the V2 point get response schema does not match the V1 collection get response (and vice versa with V1 point, V2 collection).

We can add another linter rule to specifically point out that endpoints which only differ by the scope (e.g. subscription vs. tenant scope) must have the same response schema. Explicitly showing an error for this might help clear up confusion in this scenario.

[...]

PR #765, opened 2025-01-25
Add linter rule to suggest the scope parameter

bdefoy:
Add a rule that helps disambiguate paths that share the same `/providers` suffix by suggesting the use of the `scope` parameter.

### DIFF
--- a/docs/suggest-scope-parameter.md
+++ b/docs/suggest-scope-parameter.md
@@ -1,0 +1,35 @@
+# SuggestScopeParameter
+
+## Category
+
+ARM Warning
+
+## Applies to
+
+ARM OpenAPI(swagger) specs
+
+## Related ARM Guideline Code
+
+- N/a
+
+## Description
+
+OpenAPI authors can use the `scope` path parameter to indicate that an API is applicable to various scopes (Tenant,
+Management Group, Subscription, Resource Group, etc.) rather than explicitly defining each scope in the spec.
+
+## How to fix
+
+Remove all explicitly-scoped paths that only vary in scope and create a path with the `scope` parameter.
+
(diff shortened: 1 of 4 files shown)


mikeharder (commented):
Please update `packages/rulesets/CHANGELOG.md`, similar to:

https://github.com/Azure/azure-openapi-validator/pull/762/files#diff-ff08f4e39344bc911b29791944df2958441a9fe6637e02ad9241145106267388R8

docs/suggest-scope-parameter.md
```diff
@@ -0,0 +1,35 @@
...
+
+## Description
+
+OpenAPI authors can use the `scope` path parameter to indicate that an API is applicable to various scopes (Tenant,
+Management Group, Subscription, Resource Group, etc.) rather than explicitly defining each scope in the spec.
+
+## How to fix
+
+Remove all explicitly-scoped paths that only vary in scope and create a path with the `scope` parameter.
+
```
rkmanda, line 23: This may not be a good idea from a customer viewpoint as it makes it less obvious as to what scopes are truly supported. Lets also check how TypeSpec generates  the different API paths for this scenario and align the linter rules to work with that.

bdefoy: I don't see support in TypeSpec for the scope parameter, but I'm also not seeing how you can generate a spec with a resource at multiple scopes aside from extension resources. I might look into this a bit more and follow up

bdefoy: The scenario described in #750 is a rare case where an RP is attempting to basically duplicate endpoints. I will change the rule to look for duplicate endpoints like this.

PR #767, opened 2025-02-03
Fix RequiredPropertiesMissingInResourceModel to account get and list …

AkhilaIlla:
The `RequiredPropertiesMissingInResourceModel` rule did not account for GET and LIST calls. This PR resolves the issue and also updates the `getProperties` helper method to include the LIST call pattern when fetching properties.

### DIFF
--- a/packages/rulesets/src/native/tests/resources/ext-resource-validation.json
+++ b/packages/rulesets/src/native/tests/resources/ext-resource-validation.json
@@ -94,7 +94,7 @@
       }
     },
     "/providers/Microsoft.Cache/locations/{location}/publishers": {
-      "put": {
+      "get": {
         "tags": [
           "extensions"
         ],
@@ -119,7 +119,7 @@
       }
     },
     "/providers/Microsoft.Cache/locations/{location}/Connectors": {
-      "put": {
+      "get": {
         "tags": [
           "extensions"
         ],
--- a/packages/rulesets/src/native/tests/resources/ext-resource-validation.json
+++ b/packages/rulesets/src/native/tests/resources/ext-resource-validation.json
@@ -92,6 +92,56 @@
           }
         }
       }
+    },
+    "/providers/Microsoft.Cache/locations/{location}/publishers": {
+      "put": {
+        "tags": [
+          "extensions"
+        ],
+        "operationId": "ExtensionPublisher_List",
+        "description": "Gets all Extension publishers based on the location",
+        "parameters": [
+          {
+            "$ref": "#/parameters/ApiVersionParameter"
+          },
+          {
+            "$ref": "#/parameters/SubscriptionIdParamterer"
+          }
+        ],
+        "responses": {
+          "200": {
+            "description": "OK",
+            "schema": {
+              "$ref": "#/definitions/ExtensionPublisherListResult"
+            }
+          }
+        }
+      }
+    },
+    "/providers/Microsoft.Cache/locations/{location}/Connectors": {
+      "put": {
+        "tags": [
+          "extensions"
+        ],
+        "operationId": "ExtensionPublisher_List",
+        "description": "Gets all Extension publishers based on the location",
+        "parameters": [
+          {
+            "$ref": "#/parameters/ApiVersionParameter"
+          },
+          {
+            "$ref": "#/parameters/SubscriptionIdParamterer"
+          }
+        ],
+        "responses": {
+          "200": {
+            "description": "OK",
+            "schema": {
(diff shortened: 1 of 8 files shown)


packages/rulesets/src/native/utilities/swagger-workspace.ts
```diff
@@ -83,18 +82,25 @@ export namespace Workspace {
     if (!source || !source.value) {
       return []
     }
-    let result:{[key:string]:EnhancedSchema} = {}
-    const model = source.value
+    let result: { [key: string]: EnhancedSchema } = {}
+    let model = source.value
+    // for list call, the properties are in the value.items.$ref
+    if (model.properties && model.properties.value && model.properties.value.items && model.properties.value.items.$ref) {
```
bdefoy, line 88: I think you can simplify it like this
```suggestion
    if (model?.properties?.value?.items?.$ref) {
```

packages/rulesets/src/native/tests/resources/ext-resource-validation.json
```diff
@@ -92,6 +92,56 @@
           }
         }
       }
+    },
+    "/providers/Microsoft.Cache/locations/{location}/publishers": {
+      "put": {
```
bdefoy, line 97: is this supposed to be get?

packages/rulesets/src/native/utilities/swagger-workspace.ts
AkhilaIlla: Ah good point.
thanks.

packages/rulesets/src/native/tests/resources/ext-resource-validation.json
AkhilaIlla: once of the test was to check if it is able to fetch prop's of GET
and the other test is to check how it functions when there are nested properties.

Updating to GET though.

bdefoy (approved):

rkmanda (approved):
:shipit:

FumingZhang:
Hey team, I think the latest release of openapi-validator includes this change, which caused our swagger changes to be scanned with this error.

image
https://github.com/Azure/azure-rest-api-specs/pull/33237/checks?check_run_id=38938344599

I'd like to get your ideas on how to fix errors like the first one here. The model wrapped by this list result is a model ([OperationStatusResult](https://github.com/Azure/azure-rest-api-specs/blob/main/specification/common-types/resource-management/v5/types.json#L270)) defined in common types. Should I fix the definition in the common type like this?
https://github.com/Azure/azure-rest-api-specs/pull/33239

mikeharder:
> Hey team, I think the latest release of openapi-validator includes this change, which caused our swagger changes to be scanned with this error.

Another PR hitting this:

https://github.com/Azure/azure-rest-api-specs/pull/33247/checks?check_run_id=38926358488

audunn:
In https://github.com/Azure/azure-rest-api-specs/pull/33247/checks?check_run_id=38926358488 we are following this  https://armwiki.azurewebsites.net/api_contracts/UsagesAPIContract.html?q=usages 

Looks like that contract is incompatable with this rule.   
So not sure if we should follow contract or change the spec to fix the `RequiredPropertiesMissingInResourceModel` rule. 
 
Perhaps a suppressions is the right move but then again have a suspicion this may affect more RPs. Not sure if putting suppressions is right if that is the case.

bdefoy:
@rkmanda confirmed that we are going to need to make an exception for `OperationStatusResult` as it does not have `type` per the RPC: https://github.com/cloud-and-ai-microsoft/resource-provider-contract/blob/master/v1.0/async-api-reference.md#azure-asyncoperation-resource-format

Fat-Salty-fish:
Hi team, could you please provide what's the status of this fixing? I'm still facing this issue: https://github.com/Azure/azure-rest-api-specs/actions/runs/17994009240

mikeharder:
> Hi team, could you please provide what's the status of this fixing? I'm still facing this issue: https://github.com/Azure/azure-rest-api-specs/actions/runs/17994009240

Why do you think you are hitting this issue?  Looking at your spec, the error reported appears valid.  `ResourceProviderOperation` contains property `name` but not `id` or `type`.

https://github.com/Azure/azure-rest-api-specs/blob/f6626a0b31b9715a61ded56336738a4430fe74f0/specification/desktopvirtualization/resource-manager/Microsoft.DesktopVirtualization/DesktopVirtualization/stable/2025-10-10/desktopvirtualization.json#L4287

PR #769, opened 2025-02-20
Improve ArmResourcePropertiesBag rule

tejaswiMinnu:
Added a new condition to proxy resources with basic top-level properties.
Included more top-level properties to the existing tracked resources list.

Workitems 
[Tracked resource](https://msazure.visualstudio.com/One/_workitems/edit/29144800)
[Proxy Resources](https://msazure.visualstudio.com/One/_workitems/edit/29253492)


rkmanda (approved):
:shipit:

PR #772, opened 2025-03-11
Add rule RPC-Async-V1-06 to staging

tejaswiMinnu:
Updated rule RPC-Async-V1-06 to staging only

### DIFF
--- a/packages/rulesets/generated/spectral/az-arm.js
+++ b/packages/rulesets/generated/spectral/az-arm.js
@@ -3328,6 +3328,7 @@ const ruleset = {
             description: "All long-running operations must include an `Azure-AsyncOperation` response header.",
             message: "{{description}}",
             severity: "error",
+            stagingOnly: true,
             formats: [oas2],
             given: ["$[paths,'x-ms-paths'].*.*[?(@property === 'x-ms-long-running-operation' && @ === true)]^.responses.*"],
             then: {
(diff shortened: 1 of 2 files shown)


packages/rulesets/generated/spectral/az-arm.js
```diff
@@ -3328,6 +3328,7 @@ const ruleset = {
             description: "All long-running operations must include an `Azure-AsyncOperation` response header.",
             message: "{{description}}",
             severity: "error",
+            stagingOnly: true,
```
mikeharder, line 3331: Do you have a way of testing staging rules?  The previous LintDiff "staging" pipeline has been deprecated.  We may want to find a different way to test new rules against organic PRs.

tejaswiMinnu: I don't think we have another option, so I'll discuss it with Roopesh.
I've added this to staging for now to prevent it from accidentally being shipped to production. We are waiting for some changes in TypeSpec, which Mark Cowlishaw needs to review and respond to.

bdefoy (approved):

packages/rulesets/generated/spectral/az-arm.js
rkmanda: @mikeharder Would adding this flag atleast ensure that the rule is not run in the production swagger lintdiff pipeline? For now the main thing we want to achieve is to not accidentally ship this rule until a change is made in TypeSpec for which we are following up with Mark Cowlishaw.

rkmanda (approved):

packages/rulesets/generated/spectral/az-arm.js
mikeharder: Yes, the flag should disable the rule in prod.  But, why merge this PR to main at all until TypeSpec is ready?

PR #785, opened 2025-08-20
Create staging lint checks workflow

AkhilaIlla:
This PR introduces a workflow to validate linter rules that are under development. It ensures these rules are tested against the specifications, allowing us to build confidence in them before merging and releasing to production.

How the staging-lint-checks workflow runs selected rules

Root workflow: staging-lint-checks.yaml
Triggers: pull_request events (opened, edited, synchronize, labeled, unlabeled)
Flow:
Checks out this repo and sets up Node 18
Derives RULE_NAMES from PR labels (test-<Rule>) or a “rules: A,B” line in the PR body
Checks out azure-rest-api-specs into specs/
Installs minimal deps for the runner
Runs the script [run-selected-rules.js](vscode-file://vscode-app/c:/Users/akhilailla/AppData/Local/Programs/Microsoft%20VS%20Code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) with env:
RULE_NAMES, SPEC_ROOT=specs/, FAIL_ON_ERRORS, OUTPUT_FILE
The runner filters to JSON under specification/<rp>/resource-manager/stable for key RPs and executes only the selected Spectral rules
Uploads artifacts/linter-findings.txt as a build artifact

Rules:PutResponseCodes

Latest workflow run: https://github.com/Azure/azure-openapi-validator/actions/runs/18791474287/job/53622542134
Output: 
[linter-findings.txt](https://github.com/user-attachments/files/23135320/linter-findings.txt)

### DIFF
--- a/.github/workflows/README-staging-lint-checks.md
+++ b/.github/workflows/README-staging-lint-checks.md
@@ -52,7 +52,7 @@ env:
   SPEC_REPO: Azure/azure-rest-api-specs
   FAIL_ON_ERRORS: "false"
   MAX_FILES: "100"
-  ALLOWED_RPS: "network,compute,monitor,sql,hdinsight,resource,storage"
+  ALLOWED_RPS: "compute,monitor,sql,hdinsight,network,resource,storage"
 ```
 
 - **SPEC_REPO**: Source repository for OpenAPI specifications


.github/workflows/staging-lint-checks.yaml
```diff
@@ -0,0 +1,83 @@
...
+        uses: actions/checkout@v4
+        with:
+          repository: ${{ env.SPEC_REPO }}
+          path: ${{ env.SPEC_CHECKOUT_PATH }}
+
+      - name: Install runner dependencies
+        run: |
+          npm --version
+          npm init -y
+          npm i @microsoft.azure/openapi-validator-rulesets @stoplight/spectral-core @stoplight/json-ref-resolver js-yaml
```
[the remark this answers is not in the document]
mikeharder: I agree with Daniel.  My idea was this workflow would run on PRs to this repo, and test the build of aov produced from your PR, against specs from main.

.github/scripts/package.json
```diff
@@ -0,0 +1,17 @@
...
+  "scripts": {
+    "test": "jest --runInBand"
+  },
+  "devDependencies": {
+    "jest": "^29.7.0"
+  },
+  "dependencies": {
+    "@microsoft.azure/openapi-validator-rulesets": "^2.1.7",
+    "@stoplight/spectral-core": "^1.18.3",
+    "@stoplight/json-ref-resolver": "^3.1.6",
```
mikeharder, line 14:
```suggestion
```

These shouldn't be necessary, if we build aov from the repo and then run it.

.github/scripts/package.json
```diff
@@ -0,0 +1,17 @@
+{
+  "name": "validator-runner-tests",
+  "private": true,
+  "type": "commonjs",
+  "scripts": {
+    "test": "jest --runInBand"
+  },
+  "devDependencies": {
+    "jest": "^29.7.0"
```
mikeharder, line 9:
```suggestion
    "vitest": "^3.1.2"
```

We are using vitest instead of jest for all new tests going forward (and will be converting older projects to vitest eventually).

.github/scripts/run-selected-rules.js
```diff
@@ -0,0 +1,318 @@
+"use strict"
```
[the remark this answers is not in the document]
mikeharder: Assuming we still have some script code to run autorest on specs in a loop, this tool should be refactored to be similar to tool "spec-model" in the specs repo (but ignore the "shared" folder):

1. "bin" entrypoint in `.github/package.json`

https://github.com/Azure/azure-rest-api-specs/blob/09695f5a97dfcf201766ea2f2e7e3b53aec49144/.github/shared/package.json#L30

2. Small script under `.github/cmd`

https://github.com/Azure/azure-rest-api-specs/blob/main/.github/shared/cmd/spec-model.js

3. Source code (unit testable) under `.github/src`

https://github.com/Azure/azure-rest-api-specs/blob/main/.github/shared/src/spec-model.js

mikeharder:
@AkhilaIlla: @danieljurek and I were discussing, and we think this might work better if we put the workflow in the specs repo instead.  It would work like this:

1. Trigger WF in specs repo manually
2. When triggering, specify parameters like "PR number to aov", "Rules to test", "specs to test" etc
3. WF downloads private builds from aov PR artifacts (or clones and builds locally)
4. WF runs rule against selected specs, using private build of aov

This has several benefits:

1. We can test specs in any branch in specs or specs-pr.  If the WF lives in aov repo, can only test public specs repo.
2. specs repo already has a lot of shared code and example WFs
3. WF could be slightly tweaked, to allow testing published versions of aov in addition to private builds
4. Long-term, I'd like to move aov into the specs repo, so we don't need to publish/consume at all

.github/scripts/extract-rule-names.js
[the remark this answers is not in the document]
AkhilaIlla: updated to maintain consistency.

.github/package.json
```diff
@@ -0,0 +1,11 @@
+{
+  "name": "validator-runner-tests",
+  "private": true,
+  "type": "commonjs",
```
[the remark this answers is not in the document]
AkhilaIlla: updated

.github/package.json
```diff
@@ -0,0 +1,11 @@
...
+  "scripts": {
+    "test": "vitest run --pool=forks --poolOptions.forks.singleFork"
```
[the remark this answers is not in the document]
AkhilaIlla: updated

.github/tests/extract-rule-names.test.js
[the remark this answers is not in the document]
AkhilaIlla: moved.

.github/scripts/run-autorest-selected-rules.js
```diff
@@ -0,0 +1,158 @@
...
+const fs = require('fs');
+const path = require('path');
+const { spawnSync } = require('child_process');
+
+function env(name, def='') { return (process.env[name]||'').trim() || def; }
+
+/**
+ * Check if a file path is a valid ARM spec
+ */
+function isArmSpec(filePath) {
```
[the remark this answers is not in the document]
AkhilaIlla: This function checks if the spec is a standard release in addition to what you mentioned.

.github/scripts/run-autorest-selected-rules.js
```diff
@@ -0,0 +1,158 @@
...
+    const fp = path.join(dir, e.name);
+    if (e.isDirectory()) { const r = walk(fp, cb); if (r==='STOP') return 'STOP'; }
+    else if (e.isFile()) { const r = cb(fp); if (r==='STOP') return 'STOP'; }
+  }
+}
+
+/**
+ * Collect ARM spec files from specified resource providers
+ */
+function collectSpecs(specRoot, allowedRPs, maxFiles) {
```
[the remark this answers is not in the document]
AkhilaIlla: I updated to use what you suggested. However, any change to the folder structure, naming will make the pipeline to fail.
Its different if both are in the same repo, update to one will reflect in the other and we are not sure when will things combine into one. I would have argued keeping it the same, bur went ahead and updated.

@mikeharder

.github/scripts/run-autorest-selected-rules.js
```diff
@@ -0,0 +1,158 @@
...
+  const rel = path.relative(specRoot, specPath).replace(/\\/g,'/');
+  const args = [ 'exec','--','autorest',
+    '--level=warning','--v3','--spectral','--azure-validator',
+    '--semantic-validator=false','--model-validator=false',
+    '--openapi-type=arm','--openapi-subtype=arm','--message-format=json',
+    // Pass selected rules down to the validator so only those rules execute inside the plugins.
+    `--selected-rules=${selectedRules.join(',')}`,
+    `--use=${path.join(process.cwd(),'packages','azure-openapi-validator','autorest')}`,
+    `--input-file=${specPath}` ];
+  const res = spawnSync('npm', args, {encoding:'utf8', shell:true});
```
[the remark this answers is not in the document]
AkhilaIlla: same as other.

.github/workflows/test/extract-rule-names.test.js
```diff
@@ -0,0 +1,170 @@
...
+        description: 'returns empty array when no pull request in context',
+        context: { payload: {} },
+        expected: []
+      }
+    ])('$description', ({ context, expected }) => {
+      const result = extractRuleNames(context)
+      expect(result).toEqual(expected)
+    })
+  })
+})
```
mikeharder:
```suggestion
})

```

.github/workflows/test/fixtures/lro-post-demo.json
```diff
@@ -0,0 +1,59 @@
...
+            "description": "accepted"
+          },
+          "default": {
+            "description": "Error"
+          }
+        }
+      }
+    }
+  }
+}
```
mikeharder:
```suggestion
}

```

.github/workflows/test/runner.test.js
```diff
@@ -0,0 +1,109 @@
...
+    const res = run({ SPEC_ROOT: tmpRoot, RULE_NAMES: 'PostResponseCodes,DeleteMustNotHaveRequestBody', OUTPUT_FILE: outFile, FAIL_ON_ERRORS: 'true' })
+    
+    // IN_TEST gating keeps exit code 0
+    expect(res.status).toBe(0)
+    const out = fs.readFileSync(outFile, 'utf8')
+    expect(out).toMatch(/ERROR\s*\|\s*PostResponseCodes/i)
+    expect(out).toMatch(/ERROR\s*\|\s*DeleteMustNotHaveRequestBody/i)
+    expect(out).toMatch(/Files scanned:\s*\d+/i)
+  })
+})
```
mikeharder:
```suggestion
})

```

.github/workflows/staging-lint-checks.yaml
```diff
@@ -0,0 +1,54 @@
...
+        uses: actions/setup-node@v4
+        with:
+          node-version: 22
+          
+      - name: Checkout full specs repository
+        uses: actions/checkout@v4
+        with:
+          repository: ${{ env.SPEC_REPO }}
+          path: ${{ env.SPEC_CHECKOUT_PATH }}
+          fetch-depth: 0
```
mikeharder:
```suggestion
```

`fetch-depth:0` is expensive since it checks out all history.  Can this use the default `fetch-depth:1` which is faster?

AkhilaIlla: updated to sparse checkout

.github/workflows/staging-lint-checks.yaml
```diff
@@ -0,0 +1,71 @@
...
+env:
+  SPEC_REPO: Azure/azure-rest-api-specs
+  SPEC_CHECKOUT_PATH: specs
+  FAIL_ON_ERRORS: "false" # Set to "true" to fail the job when error findings appear
+  MAX_FILES: "100"
+  ALLOWED_RPS: "compute,monitor,sql,hdinsight,network,resource,storage"
+
+jobs:
+  run-selected-rules:
+    runs-on: ubuntu-latest
```
mikeharder:
```suggestion
    runs-on: ubuntu-24.04
```

More reliable to pin to a version, and upgrade once every 2-4 years when we want.

.github/workflows/staging-lint-checks.yaml
```diff
@@ -0,0 +1,74 @@
...
+      - name: Build validator (local changes)
+        run: |
+          node common/scripts/install-run-rush.js build -t @microsoft.azure/openapi-validator
+
+      - name: Extract rules and run AutoRest validation
+        uses: actions/github-script@v7
+        with:
+          script: |
+            const { runInGitHubActions } = await import('${{ github.workspace }}/.github/workflows/src/extract-rule-names-and-run-validation.js');
+            await runInGitHubActions(context, core, process.env);
```
mikeharder:
```suggestion
            await runInGitHubActions(context, core);
```

No need to pass this, the other file can just access `process.env`.

.github/package-lock.json
mikeharder: package-lock.json required for repeatable builds (eg run unit tests in .github)

.github/workflows/github-test.yaml
```diff
@@ -68,6 +68,9 @@ jobs:
       - name: Install dev deps
         run: npm ci
 
+      - run: npm run test:ci
```
mikeharder: Temp change to run tests even with lint failing.  Revert before merging.

.github/workflows/src/extract-rule-names-and-run-validation.js
```diff
@@ -0,0 +1,372 @@
...
+ * the selected rules against Azure REST API specs.
+ * 
+ * Usage in GitHub Actions:
+ * Uses github-script action with direct access to context and core
+ */
+
+import fs from 'fs';
+import path from 'path';
+import { fileURLToPath } from 'url';
+import { spawnSync } from 'node:child_process';
```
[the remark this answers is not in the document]
AkhilaIlla: same

.github/workflows/staging-lint-checks.yaml
AkhilaIlla: makes sense.

.github/workflows/src/extract-rule-names-and-run-validation.js
```diff
@@ -0,0 +1,372 @@
...
+    '--level=warning','--v3','--spectral','--azure-validator',
+    '--semantic-validator=false','--model-validator=false',
+    '--openapi-type=arm','--openapi-subtype=arm','--message-format=json',
+    // Pass selected rules down to the validator so only those rules execute inside the plugins.
+    `--selected-rules=${selectedRules.join(',')}`,
+    `--use=${path.join(repoRoot, 'packages', 'azure-openapi-validator', 'autorest')}`,
+    `--input-file=${specPath}`
+  ];
+  
+  const result = spawnSync('npm', args, {encoding:'utf8', shell:true});
```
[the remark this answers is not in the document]
mikeharder: I can convert this to use our exec helper, once the other errors are fixed.

mikeharder (approved):

.github/workflows/src/extract-rule-names-and-run-validation.js
```diff
@@ -0,0 +1,404 @@
...
+  return ruleLine
+    .split(/[\n,]/)
+    .map((s) => s.trim())
+    .filter(Boolean);
+}
+
+/**
+ * Extract and combine rule names from GitHub context
+ * Combines rules from both labels and body, removing duplicates
+ * @param {import('@actions/github-script').AsyncFunctionArguments['context']} context  - GitHub Actions context object
```
mikeharder:
```suggestion
 * @param {import('@actions/github-script').AsyncFunctionArguments['context']} context - GitHub Actions context object
```

.github/workflows/src/extract-rule-names-and-run-validation.js
```diff
@@ -0,0 +1,404 @@
...
+    core.setFailed(`Script execution failed: ${errorMessage}`);
+    throw error;
+  }
+}
+
+/**
+ * Core validation logic
+ * @param {string[]} selectedRules - Array of rule names to validate
+ * @param {any} env - Environment variables object
+ * @param {import('@actions/github-script').AsyncFunctionArguments['core'] | null} core  - GitHub Actions core object (optional)
```
mikeharder:
```suggestion
 * @param {import('@actions/github-script').AsyncFunctionArguments['core'] | null} core - GitHub Actions core object (optional)
```

PR #787, opened 2025-09-03
[getProperties] Use "items" schema if it has "properties" or "allOf"

### LINKED ISSUE #786: [RequiredPropertiesMissingInResourceModel] False positive if array element model has no "properties" property
In this PR, `RequiredPropertiesMissingInResourceModel` failed with what appears to be a false positive:

https://github.com/Azure/azure-rest-api-specs/actions/runs/17415866870

I believe the bug is here:

https://github.com/Azure/azure-openapi-validator/blob/2787ae4dbb15b534d4a73bcc2cfe1e5772010f5e/packages/rulesets/src/native/utilities/swagger-workspace.ts#L94-L99

Image

In this spec, `referenceSchema.value.properties` is undefined, because the value is just `allOf` two other models, with no `properties` of its own:

```
    "ContainerGroupProfile": {
      "description": "A container group profile.",
      "allOf": [
        {
          "$ref": "#/definitions/Resource"
        },
        {
          "$ref": "#/definitions/ContainerGroupProfileProperties"
        }
      ]
    },
```

https://github.com/Azure/azure-rest-api-specs/blob/2c1d4b6d885719edf0df8de4f289c78c6238d4a3/specification/containerinstance/resource-manager/Microsoft.ContainerInstance/stable/2025-09-01/containerInstance.json#L1758-L1768

mikeharder:
- Fixes issue, where schemas with "allOf" but no "properties" were ignored
- Fixes #786

## TypeSpec
TypeSpec doesn't generate multiple allOf, at least not in widget.  Instead, it puts the extra properties under a "properties" bag.

https://github.com/Azure/azure-rest-api-specs/blob/main/specification/widget/resource-manager/Microsoft.Widget/Widget/stable/2021-11-01/widget.json
  
TSP generates this:

```json
    "Employee": {
      "type": "object",
      "description": "Employee resource",
      "properties": {
        "properties": {
          "$ref": "#/definitions/EmployeeProperties",
          "description": "The resource-specific properties for this resource."
        }
      },
      "allOf": [
        {
          "$ref": "../../../../../../common-types/resource-management/v5/types.json#/definitions/TrackedResource"
        }
      ]
    },
```

instead of this:

```json
    "Employee": {
      "type": "object",
      "description": "Employee resource",
      "allOf": [
        {
          "$ref": "../../../../../../common-types/resource-management/v5/types.json#/definitions/TrackedResource"
        },
        {
          "$ref": "#/definitions/EmployeeProperties"
        }, 
      ]
    },
```

This is primarily an implementation detail / design decision in the TSP autorest emitter, but in theory they could emit multiple allOf instead.


packages/rulesets/src/native/utilities/swagger-workspace.ts
```diff
@@ -93,7 +93,7 @@ export namespace Workspace {
     }
     else if(reference){
       const referenceSchema = resolveRef(createEnhancedSchema(model.properties.value.items, source.file), inventory)
-      if (referenceSchema && referenceSchema.value && referenceSchema.value.properties) {
+      if (referenceSchema && referenceSchema.value && (referenceSchema.value.properties || referenceSchema.value.allOf)) {
```
mikeharder, line 96: Below, the code merges the contents of `model.properties` and `model.allOf`, so this conditional should be true if **either** is defined on `referenceSchema.value`.

PR #749, opened 2024-10-14
Added new rule LroAzureAsyncOperationHeader

tejaswiMinnu:
LroAzureAsyncOperationHeader rule checks "Azure-AsyncOperation" header must be added for all async operations.

This checks for the absence of headers, including the "Azure-AsyncOperation" header.

[WorkItem](https://msazure.visualstudio.com/One/_workitems/edit/29424555)

PR #788, opened 2025-09-04
Revert "Added new rule LroAzureAsyncOperationHeader (#749)"

mikeharder:
- Reverts commit 2bff5308a52f642959daf0a9b7a10823b05a8c8c
- Rule was blocking too many PRs with suspected false positives


packages/rulesets/CHANGELOG.md
```diff
@@ -1,10 +1,17 @@
...
+### Patches
+
+- Removed rule LroAzureAsyncOperationHeader
+
 ## 2.1.8
 
 ### Patches
 
 - Modified the rule TenantLevelAPIsNotAllowed RPC-Uri-V1-11 from error to warning
+- Added rule LroAzureAsyncOperationHeader
```
AkhilaIlla, line 14: Well, I think we should make sure these are accurate before we release.

I remember there was a process where the version bump added all the changelogs that were supposed to be under that version.

PR #790, opened 2025-09-10
[CODEOWNERS] Add @raosuhas

mikeharder:


.github/CODEOWNERS
```diff
@@ -1,16 +1,15 @@
 # About this file:
 # https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners
 
-# @konrad-jamrozik is the primary owner of this repository.
-# @mikeharder and @weshaggard are secondary owners.
-/* @konrad-jamrozik @mikeharder @weshaggard
+# @mikeharder is the primary owner of this repository.
+# @raosuhas and @weshaggard are secondary owners.
+/* @mikeharder @raosuhas @weshaggard
```
mikeharder, line 6:
```suggestion
* @mikeharder @raosuhas @weshaggard
```

.github/CODEOWNERS
```diff
@@ -1,16 +1,15 @@
 # About this file:
 # https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners
 
-# @konrad-jamrozik is the primary owner of this repository.
-# @mikeharder and @weshaggard are secondary owners.
-/* @konrad-jamrozik @mikeharder @weshaggard
+# @mikeharder is the primary owner of this repository.
+# @raosuhas and @weshaggard are secondary owners.
+* @mikeharder @raosuhas @weshaggard
```
mikeharder, line 6: The change from `/*` to `*` is intentional.  The former only matches files directly in root, while the latter matches all files in the repo (not matched by a later rule).  I think the intention was to match all files, and the use of `/*` was a bug.

PR #800, opened 2025-09-25
Add new rule RPC-v1-13 to mandate  x-ms-secrets annotation for properties containing secrets for PUT/GET responses

tejaswiMinnu:


mikeharder (changes requested):
Please update changelog and package.json.  Since this is a new rule (new feature), bump the minor.

tejaswiMinnu:
> Please update changelog and package.json. Since this is a new rule (new feature), bump the minor.

Thanks Mike, updated

docs/xms-secret-in-response.md
```diff
@@ -0,0 +1,85 @@
...
+
+## Applies to
+
+ARM OpenAPI(swagger) specs
+
+## Related ARM Guideline Code
+
+- RPC-Put-V1-13
+
+## Output Message
```
AkhilaIlla, line 15: I dont think the standard template has this field.
[https://github.com/Azure/azure-openapi-validator/blob/main/CONTRIBUTING.md#spectral-rule](https://github.com/Azure/azure-openapi-validator/blob/main/CONTRIBUTING.md#spectral-rule)

packages/rulesets/CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Change Log - @microsoft.azure/openapi-validator-rulesets
 
+## 2.2.0
```
AkhilaIlla, line 3: So, for any new change we add it under this version or add a new version?

How will we know if a new version is needed or not?

packages/rulesets/src/spectral/az-arm.ts
```diff
@@ -683,6 +684,18 @@ const ruleset: any = {
         function: withXmsResource,
       },
     },
+    // RPC Code: RPC-Put-V1-13
+    XMSSecretInResponse: {
+      rpcGuidelineCode: "RPC-Put-V1-13",
+      description: `Properties contains secret keyword and does not have 'x-ms-secret' annotation. To ensure security, must add the 'x-ms-secret' annotation to this property.`,
```
AkhilaIlla, line 690: >`Properties contains secret keyword and does not have 'x-ms-secret' annotation. To ensure security, must add the 'x-ms-secret' annotation to this property.`, [](http://example.com/codeflow?start=19&length=157)

I think description should be more generic explaining what the rule is (like the one you have in the docs) and not like an error/output message.

packages/rulesets/CHANGELOG.md
mikeharder: > So, for any new change we add it under this version or add a new version?  How will we know if a new version is needed or not?

The goal is that main should be "always shippable", if there are any customer-facing changes in the branch.  Here's what this means in practice, when creating a PR to this repo:

1. Look at the current version in main/package.json, and the latest version published to npmjs.com.  They might be the same, or the version in main might have already been incremented.

2. Determine the customer-facing impact of your change.
    1. If breaking change, must bump the major version
    2. If new feature (e.g. new rule), must bump the minor 
    3. If bug fix (e.g. fixing bug in rule, bumping runtime dep version), must bump the patch
    4. If no customer impact (e.g. bumpting dev dep version), don't change the version at all.

3. Given the results of steps 1 and 2, determine if you need to actually bump the version, or can "batch" your change with another unrelased change.

Our goal is to ship as often as possible (e.g. after every bug fix or new feature), since publishing and updating specs is now relatively cheap.  However, it's also fine to batch a few changes together into a single release, if it's more efficient.

packages/rulesets/package.json
```diff
@@ -86,4 +86,4 @@
       "^@stoplight/spectral-ruleset-bundler/(.*)$": "<rootDir>/node_modules/@stoplight/spectral-ruleset-bundler/dist/$1"
     }
   }
-}
+}
```
mikeharder, line 89:
```suggestion
}

```

Add trailing newline

packages/rulesets/src/spectral/az-arm.ts
tejaswiMinnu: Modified

docs/xms-secret-in-response.md
tejaswiMinnu: Removed

docs/xms-secret-in-response.md
```diff
@@ -0,0 +1,85 @@
...
+
+Properties contains secret keyword and does not have 'x-ms-secret' annotation. To ensure security, must add the 'x-ms-secret' annotation to this property.
+
+## Description
+
+When defining the response model for an ARM PUT or GET operation, any property that contains sensitive information (such as passwords, keys, tokens, credentials, or other secrets) must include the `"x-ms-secret": true` annotation. This ensures that secrets are properly identified and handled according to ARM security guidelines.
+
+## How to fix the violation
+
+To fix this violation, review the response model for your ARM PUT/GET operation and identify all properties that store secrets (such as `password`, `key`, `access`, `credentials`, `token`, `secret`, `auth`). For each of these properties, add the `"x-ms-secret": true` annotation to ensure they are properly marked as secrets. This helps protect sensitive information and complies with ARM guidelines.
```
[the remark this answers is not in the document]
AkhilaIlla: I dont have access. However what was the reason to just consider PUT in addition to GET and not POST/DELETE?

packages/rulesets/generated/spectral/az-arm.js
```diff
@@ -3234,6 +3234,31 @@ const xmsPageableForListCalls = (swaggerObj, _opts, paths) => {
...
+    }
+    const secretKeywords = ["access", "credentials", "secret", "password", "key", "token", "auth", "connection"];
+    const path = ctx.path || [];
+    const errors = [];
+    for (const prpName of Object.keys(properties)) {
+        if (prpName === "properties" && typeof properties[prpName] === "object") {
+            errors.push(...XMSSecretInResponse(properties[prpName], _opts, { ...ctx, path: [...path, prpName] }));
+        }
+        else {
+            if (secretKeywords.some((keyword) => prpName.toLowerCase().includes(keyword))) {
```
rajeshchadadev, line 3249: Did we consider false positives if property name contains below example substrings which are not secrets/credentials:
1. "keyvaultid" instead of "key" and keyvault key resource info
2. "accesspolicies" instead of "access"
3. "secretreferenceid" instead of "secret"
4. "storageKeyType" instead of "key"
5. "authenticationType" instead of "auth"

Also did we consider false-negative scenarios for patterns like:
1. "credential" instead of "credentials"

docs/rules.md
```diff
@@ -1446,3 +1446,9 @@ Please refer to [xms-paths-must-overload-paths.md](./xms-paths-must-overload-pat
 The 200 response model for an ARM PUT operation must have x-ms-azure-resource extension set to true in its hierarchy. Operation: '{0}' Model: '{1}'.
 
 Please refer to [xms-resource-in-put-response.md](./xms-resource-in-put-response.md) for details.
+
+### XmsResourceInPutResponse
+
+When defining the response model for an ARM PUT or GET operation, any property that contains sensitive information (such as passwords, keys, tokens, credentials, or other secrets) must include the `"x-ms-secret": true` annotation. This ensures that secrets are properly identified and handled according to ARM security guidelines.
```
AkhilaIlla, line 1452: >PUT or GET  [](http://example.com/codeflow?start=44&length=11)

include POST

mikeharder:
> Please release this rule to the pipeline only after the associated task has been merged: [Azure/typespec-azure#3403](https://github.com/Azure/typespec-azure/issues/3403). The tentative timeline for this is 11/11.

@tejaswiMinnu: main should always be "ready to release", so this should not have been merged until you were ready for it to go live in the specs repo.  If nobody needs to release before 11/11, this can stay.  Else, it will need to be reverted and re-added later.

docs/rules.md
timotheeguerin: looking to add this rule to TypeSpec and this is flagging a **lot** of things https://github.com/Azure/typespec-azure/actions/runs/18606539985/job/53057024914
with the vast majority looking like false positives

Image

Keyvault for example has 44 errors right now mostly about properties called `keyName`

should this logic be tweaked a little this feels quite agressive otherwise.

rajeshchadadev: DevSec team has anticipated this when we reviewed the PR which can be found under the comment: https://github.com/Azure/azure-openapi-validator/pull/800#discussion_r2409835336

PR #805, opened 2025-12-01
Add new rule RPC-v1-13 to mandate x-ms-secrets annotation for properties containing secrets

tejaswiMinnu:
This reverts commit a73dd472198899cc703fb764c822721627203689.

When defining the response model for an ARM PUT or GET operation, any property that contains sensitive information (such as passwords, keys, tokens, credentials, or other secrets) must include the "x-ms-secret": true annotation. This ensures that secrets are properly identified and handled according to ARM security guidelines.

The last commit changes were based on suggestions from the typespec team. You can review the typespec PR here: https://github.com/Azure/typespec-azure/pull/3411.
- Added checks for keywords ending with sensitive secrets.
- Exclude key when used in a model that only has 1 other property value(Get rid of most key instance as its just key value pair)  

rules: XMSSecretInResponse

### DIFF
--- a/.github/package-lock.json
+++ b/.github/package-lock.json
@@ -4,6 +4,9 @@
   "requires": true,
   "packages": {
     "": {
+      "dependencies": {
+        "autorest": "3.7.2"
+      },
       "devDependencies": {
         "@actions/github-script": "github:actions/github-script",
         "@eslint/js": "^9.22.0",
@@ -2091,6 +2094,19 @@
         "@jridgewell/trace-mapping": "^0.3.31",
         "estree-walker": "^3.0.3",
         "js-tokens": "^9.0.1"
+      }
+    },
+    "node_modules/autorest": {
+      "version": "3.7.2",
+      "resolved": "https://registry.npmjs.org/autorest/-/autorest-3.7.2.tgz",
+      "integrity": "sha512-yEeF0tJjx2fROK9VWIVHKFiUSzD0cxwqnq7z+v7kIIRRZjyOM3rpBS9OPp6tQv5d3mmxPAUNh57G1ZumQNqQGg==",
+      "hasInstallScript": true,
+      "license": "MIT",
+      "bin": {
+        "autorest": "entrypoints/app.js"
+      },
+      "engines": {
+        "node": ">=12.0.0"
       }
     },
     "node_modules/balanced-match": {
--- a/.github/package.json
+++ b/.github/package.json
@@ -4,6 +4,9 @@
   "_comments": {
     "copy": "Includes subset of https://github.com/Azure/azure-rest-api-specs/blob/main/.github/package.json"
   },
+  "dependencies": {
+    "autorest": "3.7.2"
+  },
   "devDependencies": {
     "@actions/github-script": "github:actions/github-script",
     "@eslint/js": "^9.22.0",
--- a/.github/tsconfig.json
+++ b/.github/tsconfig.json
@@ -11,6 +11,7 @@
   },
   "include": [
     // Only check runtime sources.  Tests currently have too many errors.
+    "*.js",
     "**/src/**/*.js",
   ],
 }
--- a/.github/vitest.config.js
+++ b/.github/vitest.config.js
@@ -6,13 +6,15 @@ export default defineConfig({
   esbuild: {
     // Ignore tsconfig.json, since it's only used for type checking, and causes
     // a warning if vitest tries to load it
+
+    // @ts-expect-error: 'tsConfig' does not exist in type 'ESBuildOptions'
     tsConfig: false,
   },
 
   test: {
     coverage: {
       exclude: [
-        ...configDefaults.coverage.exclude,
+        ...(configDefaults.coverage.exclude ?? []),
 
         // Not worth testing CLI code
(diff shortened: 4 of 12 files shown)


AkhilaIlla (changes requested):
Test the new rule using staging linter workflow and validate if works as expected.

packages/rulesets/src/spectral/functions/xms-secret-in-response.ts
```diff
@@ -0,0 +1,52 @@
+const sensitiveKeywords = ["access", "credential", "secret", "password", "key", "token", "auth", "connection"]
+
+const excludeKeywords = ["publicKey"].map((keyword) => keyword.toLowerCase())
+
+function isPotentialSensitiveProperty(propertyName: string): boolean {
+  const lowerName = propertyName.toLowerCase()
+  return (
+    sensitiveKeywords.some((keyword) => lowerName.endsWith(keyword) && !lowerName.startsWith(keyword)) &&
```
AkhilaIlla: >lowerName.endsWith(keyword) && !lowerName.startsWith(keyword) [](http://example.com/codeflow?start=40&length=61)

why endwith & not equals? #Resolved

packages/rulesets/src/spectral/functions/xms-secret-in-response.ts
```diff
@@ -0,0 +1,52 @@
...
+  for (const prpName of propertiesKeys) {
+    if (prpName === "properties" && typeof properties[prpName] === "object") {
+      errors.push(...XMSSecretInResponse(properties[prpName], _opts, { ...ctx, path: [...path, prpName] }))
+    } else {
+      // Add all conditions for secret detection
+      if (
+        isPotentialSensitiveProperty(prpName) && // property name matches sensitive keywords
+        properties[prpName] && // property exists
+        properties[prpName]["x-ms-secret"] !== true && // not explicitly marked as secret
+        !keyValuePairCheck && // not a key-value pair key
```
AkhilaIlla: >!keyValuePairCheck && // not a key-value pair key [](http://example.com/codeflow?start=8&length=49)

what is this additional check for?

packages/rulesets/src/spectral/functions/xms-secret-in-response.ts
AkhilaIlla: nvm

packages/rulesets/src/spectral/test/xms-secret-in-response.test.ts
```diff
@@ -0,0 +1,574 @@
...
+      FooResource: {
+        "x-ms-azure-resource": true,
+        properties: {
+          password: {
+            type: "string",
+            "x-ms-secret": true,
+          },
+          token: {
+            type: "string",
+          },
```
AkhilaIlla: why is this not flagged?

packages/rulesets/src/spectral/functions/xms-secret-in-response.ts
tejaswiMinnu: Suggestion from typespec team: "Exclude key when used in a model that only has 1 other property value(Get rid of most key instance as its just key value pair)"

Without this check, we were getting a bunch of false positives. That’s why the typespec team came up with all these new checks, and now they’ve added them to their code too to avoid false positives.

packages/rulesets/src/spectral/functions/xms-secret-in-response.ts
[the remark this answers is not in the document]
tejaswiMinnu: Yep, same here—the typespec team introduced some new checks to cut down on all those false positives.

packages/rulesets/src/spectral/test/xms-secret-in-response.test.ts
```diff
@@ -0,0 +1,574 @@
...
+    expect(results[15].message).toBe(ERROR_MESSAGE.replace("{prpName}", "someconnection"))
+    expect(results[15].path.join(".")).toBe("paths./foo.put.responses.200.schema.properties.someconnection")
+    expect(results[16].message).toBe(ERROR_MESSAGE.replace("{prpName}", "somecredential"))
+    expect(results[16].path.join(".")).toBe("paths./foo.put.responses.201.schema.properties.properties.properties.somecredential")
+    expect(results[17].message).toBe(ERROR_MESSAGE.replace("{prpName}", "somesecret"))
+    expect(results[17].path.join(".")).toBe("paths./foo.put.responses.201.schema.properties.properties.properties.somesecret")
+  })
+})
+
+test("XMSSecretInResponse should find no errors with newly added typespec conditions", () => {
```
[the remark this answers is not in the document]
tejaswiMinnu: In the last commit, all checks were sourced from the typespec team. I have added a few additional tests to verify if they align with the new checks.
here is the PR - https://github.com/Azure/typespec-azure/pull/3411

packages/rulesets/src/spectral/functions/xms-secret-in-response.ts
AkhilaIlla: and why is this- Exclude key when used in a model that only has 1 other property value(Get rid of most key instance as its just key value pair)?

